PR #30, opened 2020-07-15
Fix Schema to load UserToken yaml file

sacheu:


Sources/AutorestSwift/Models/OperationGroup.swift
```diff
@@ -13,6 +13,22 @@ public struct OperationGroup: Codable {
 
     public let operations: [Operation]
 
+    enum CodingKeys: String, CodingKey {
+        case operations, key = "$key"
```
tjprescott, line 17:
```suggestion
        case operations
        case key = "$key"
```

This makes it a little cleaner and clearer.

Sources/AutorestSwift/Models/OperationGroup.swift
```diff
@@ -13,6 +13,22 @@ public struct OperationGroup: Codable {
...
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(key, forKey: .key)
+        try container.encode(operations, forKey: .operations)
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        key = try container.decode(String.self, forKey: .key)
+        operations = try container.decode([Operation].self, forKey: .operations)
+    }
```
tjprescott, line 33: Does the compiler require these? I would think that since we are providing a custom `CodingKeys` enum it should still be able to synthesize these.

Sources/AutorestSwift/main.swift
```diff
@@ -7,7 +7,11 @@ do {
     let yamlString = try String(contentsOf: url)
     let decoder = YAMLDecoder()
     let model = try decoder.decode(CodeModel.self, from: yamlString)
-    print(model)
+    let encoder = JSONEncoder()
+    encoder.outputFormatting = .prettyPrinted
+    let jsonData = try encoder.encode(model)
+
+    print(String(data: jsonData, encoding: .utf8)!)
```
tjprescott, line 14:
```suggestion
    if let jsonString = String(data: jsonData, encoding: .utf8) {
      print(jsonString)
    }
```

We haven't enabled swiftlint or other linters on this repo yet, but when we do they will complain hard about a forced unwrap (!).

Sources/AutorestSwift/Models/OperationGroup.swift
sacheu: Thanks. I test it. Yes, the compile does not require these encode/decode functions. I will remove them.

PR #31, opened 2020-07-15
Support ''allOf' property in Schema classes

sacheu:


Sources/AutorestSwift/Models/Compose.swift
```diff
@@ -0,0 +1,90 @@
+import Foundation
```
tjprescott, line 1: This needs a standard header, and it will need to credit the original article or repo IAW whatever that repos license agreement is.

tjprescott: Also, this file should probably go in a different directory from Models, which is intended only for the Code Model model types. Maybe a Util folder?

PR #37, opened 2020-07-17
Decoding Polymorphic Types

tjprescott:
A rough attempt at cobbling together polymorphic support.

Also, this updates `main.swift` so that we can see the original file in JSON format and the JSON that results after round-tripping through our Swift model. Both files are sorted and pretty printed for easier diffing.


src/AutorestSwift/Models/Protocols.swift
```diff
@@ -9,11 +9,34 @@ import Foundation
...
+        case http, amqp, mqtt, jsonrpc
+    }
+
+    // MARK: Codable
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        http = (try? container.decode(HttpParameter.self, forKey: .http)) ??
+            (try? container.decode(HttpResponse.self, forKey: .http)) ??
+            (try? container.decode(HttpModel.self, forKey: .http))
```
tjprescott, line 30: You have to be careful the way you order these. They need to be from most specific/complex to least.

sacheu: you need to add a case 'HttpWithBodyRequest' to decode Http

tjprescott: Done.

sacheu:
Thanks Travis. PR Looks good.

PR #38, opened 2020-07-17
Add http classes per schema

sacheu:


tjprescott:
Got the files. Thank @sacheu

PR #46, opened 2020-07-20
Enable extensions property

tjprescott:
This assumes `[String: Bool]` since that is what our current Swaggers are using, but really, these could also be `[String: String]` and potentially other things, so this is likely to need to be reworked.


src/AutorestSwift/Models/Parameter.swift
```diff
@@ -10,14 +10,13 @@ import Foundation
 public typealias Parameter = Compose<ParameterProperty, Value>
 
 /// A definition of an discrete input for an operation
-public struct ParameterProperty: Codable {
+public class ParameterProperty: Codable {
```
tjprescott, line 13: Must be a class to reference another property of the same type.

PR #50, opened 2020-07-20
Fix-ups

tjprescott:
With this change, I record 0 diffs on Test Swagger A and 26 diffs on Test Swagger B, which fall into exactly 2 categories.

### DIFF
--- a/src/AutorestSwift/main.swift
+++ b/src/AutorestSwift/main.swift
@@ -11,8 +11,9 @@ let afterJsonUrl = documentsUrl.appendingPathComponent("after.json")
 do {
     var yamlString = try String(contentsOf: sourceUrl)
 
-    // Work around to handle blank namespace for Language not being decoded.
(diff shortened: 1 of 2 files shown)


src/AutorestSwift/main.swift
```diff
@@ -11,8 +11,9 @@ let afterJsonUrl = documentsUrl.appendingPathComponent("after.json")
 do {
     var yamlString = try String(contentsOf: sourceUrl)
 
-    // Work around to handle blank namespace for Language not being decoded.
-    yamlString = yamlString.replacingOccurrences(of: "namespace: ''", with: "namespace: 'default'")
+    // TODO: Remove when issue (https://github.com/Azure/autorest.swift/issues/47) is fixed. Replaces empty string with a single space.
+    yamlString = yamlString.replacingOccurrences(of: ": ''", with: ": ' '")
```
tjprescott, line 15: This is a more generic fix, as this was impacting all empty strings, not just namespace, and is a bug in YAMS.

PR #55, opened 2020-07-21
Refactor language and languages

tjprescott:
This PR refactors `Language` to have a copy constructor and `Languages` to eliminate languages we don't care about and add behavior so that calling `language.swift.foo` will always succeed, copying the information in `language.default` if needed. This will allow us to write code generation using `language.swift` before the naming pass of the generator is complete.

### DIFF
--- a/src/AutorestSwift/Models/Simple/Languages.swift
+++ b/src/AutorestSwift/Models/Simple/Languages.swift
@@ -11,29 +11,31 @@ import Foundation
 public class Languages: Codable {
     public let `default`: Language
 
-    public let csharp: CSharpLanguage?
+    // these properties we can set
+    private var _swift: Language?
 
-    public let python: Language?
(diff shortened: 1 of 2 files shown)


src/AutorestSwift/Models/Simple/Languages.swift
```diff
@@ -11,29 +11,31 @@ import Foundation
...
+    }
+
+    public var objectiveC: Language?
+
+    // MARK: Codable
+
+    enum CodingKeys: String, CodingKey {
+        case `default`
+        case _swift = "swift"
+        case objectiveC
```
sacheu, line 39: the objectiveC case here is for supporting code generation for ObjectiveC?

tjprescott: Yeah, as a potential future thing.

PR #56, opened 2020-07-21
Refactor core

tjprescott:
Refactors `swift.main` to be more object oriented and facilitate work on parallel paths.

- `before.json` and `after.json` are only generated if there are differences. All files will be placed in the `Documents/generated/sdk/` folder.
- Adds a logger class instead of relying on print statements.

### DIFF
--- a/src/AutorestSwift/Code Generator/CodeNamer.swift
+++ b/src/AutorestSwift/Code Generator/CodeNamer.swift
@@ -1,0 +1,11 @@
+//
+//  CodeNamer.swift
+//  
+//
+//  Created by Travis Prescott on 7/21/20.
+//
+
+import Foundation
+
+/// Common protocol for language code namers.
+protocol CodeNamer {}
--- a/src/AutorestSwift/Code Generator/Manager.swift
+++ b/src/AutorestSwift/Code Generator/Manager.swift
@@ -1,0 +1,122 @@
+//
+//  Manager.swift
+//  
+//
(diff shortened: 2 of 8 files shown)


src/AutorestSwift/Code Generator/CodeNamer.swift
```diff
@@ -0,0 +1,11 @@
...
+//  CodeNamer.swift
+//  
+//
+//  Created by Travis Prescott on 7/21/20.
+//
+
+import Foundation
+
+/// Common protocol for language code namers.
+protocol CodeNamer {}
```
tjprescott, line 11: These are for reuse later if we decide to do, say, an ObjectiveC code namer.

src/AutorestSwift/Code Generator/Manager.swift
```diff
@@ -0,0 +1,122 @@
...
+            logger.log("Model file check: OK", level: .info)
+        }
+    }
+
+    private func ensureExists(folder: URL) {
+        let fileManager = FileManager.default
+
+        if let existing = try? folder.resourceValues(forKeys: [.isDirectoryKey]) {
+            if !existing.isDirectory! {
+                fatalError("Path exists but is not a folder!")
```
sacheu, line 111: Do we need to call logger here too?

tjprescott: Yeah, that would make sense.

PR #64, opened 2020-07-22
Use Stencil template engine to generate code for ObjectSchema and ChoiceSchema

sacheu:
Need to copy ChoiceSchema.stencil and ObjectSchema.stencil to document directory in order to work


src/AutorestSwift/Code Generator/Constants.swift
```diff
@@ -39,5 +39,4 @@ struct Constants {
     // --------------------------------------------------------------------------
```
tjprescott, line 39: We can remove this file and simply put the header in its own `.stencil` file.

src/AutorestSwift/Code Generator/Manager.swift
```diff
@@ -46,7 +46,7 @@ class Manager {
     init(withInputUrl input: URL, destinationUrl dest: URL) {
         self.inputUrl = input
         // TODO: Make this configurable
-        destinationRootUrl = dest.appendingPathComponent("generated").appendingPathComponent("sdk")
+        self.destinationRootUrl = dest.appendingPathComponent("generated").appendingPathComponent("sdk")
```
tjprescott, line 49: Is the being changed by swiftlint/swiftformat? Seems like it keeps bouncing back and forth.

src/AutorestSwift/Code Generator/SwiftGenerator.swift
```diff
@@ -34,6 +35,8 @@ class SwiftGenerator: CodeGenerator {
 
     let baseUrl: URL
 
+    let documentsUrl = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first
```
tjprescott, line 38: This shouldn't be necessary. `main.swift` passes in the baseUrl as the documents directory

src/AutorestSwift/Code Generator/SwiftGenerator.swift
```diff
@@ -46,12 +49,22 @@ class SwiftGenerator: CodeGenerator {
     /// Begin code generation process
     func generate() throws {
         // TODO: Begin code generation process
-
         try generateSchemas()
+
+        if let choices = model.schemas.choices {
```
tjprescott, line 54: Let's replace the implementation of `generateSchemas` with this logic.

src/AutorestSwift/Code Generator/SwiftGenerator.swift
```diff
@@ -92,4 +105,20 @@ class SwiftGenerator: CodeGenerator {
...
+    private func generateObject(for object: ObjectSchema) throws {
+        try renderTemplate(filename: "ObjectSchema.stencil", dictionary: ["object": object])
+    }
+
+    private func renderTemplate(filename: String, dictionary: [String: Any]) throws {
+        let fsLoader = FileSystemLoader(bundle: [Bundle.main])
+        let environment = Environment(loader: fsLoader)
+
+        var result = try environment.renderTemplate(name: filename, context: dictionary)
+        print(result)
```
tjprescott, line 122: Let's actually save these outputs to the relevant files.

src/AutorestSwift/Code Generator/SwiftGenerator.swift
```diff
@@ -92,4 +105,20 @@ class SwiftGenerator: CodeGenerator {
             try object.toFile(inFolder: modelUrl)
         }
     }
+
+    private func generateChoice(for choice: ChoiceSchema) throws {
```
tjprescott, line 109: So this will need to be a method to generate the enum file and will probably need to compose several templates.

src/AutorestSwift/Code Generator/SwiftGenerator.swift
```diff
@@ -92,4 +105,20 @@ class SwiftGenerator: CodeGenerator {
...
+        try renderTemplate(filename: "ChoiceSchema.stencil", dictionary: ["choice": choice])
+    }
+
+    private func generateObject(for object: ObjectSchema) throws {
+        try renderTemplate(filename: "ObjectSchema.stencil", dictionary: ["object": object])
+    }
+
+    private func renderTemplate(filename: String, dictionary: [String: Any]) throws {
+        let fsLoader = FileSystemLoader(bundle: [Bundle.main])
+        let environment = Environment(loader: fsLoader)
```
tjprescott, line 119: We can store the FileSystemLoader and Environment variables as private variables of the SwiftGenerator so you don't need to instantiate them repeatedly.

src/AutorestSwift/Models/Complex/ChoiceSchema.swift
```diff
@@ -61,13 +61,13 @@ public class ChoiceSchema: ValueSchema {
 extension ChoiceSchema: SnippetConvertible {
```
tjprescott, line 61: We can remove this. If you see value in a protocol to annotate which model objects have stencils associated with them, then change it to that. Otherwise we can just remove it. 

```Swift
extension ChoiceSchema: Stencilable {
  func generateTemplate() -> String {
    return renderTemplate(filename: "ChoiceSchema.stencil", dictionary: ["object": self])
  }
}
```

src/AutorestSwift/Models/Complex/ObjectSchema.swift
```diff
@@ -97,7 +97,7 @@ extension ObjectSchema: FileConvertible {
         string += "import Foundation\n\n"
```
tjprescott, line 97: Same. Let's repurpose the protocol or remove it entirely.

src/AutorestSwift/Models/Complex/SealedChoiceSchema.swift
```diff
@@ -56,13 +56,13 @@ public class SealedChoiceSchema: ValueSchema {
 extension SealedChoiceSchema: SnippetConvertible {
     func toSnippet() -> String {
         let name = self.name
```
tjprescott, line 58: Here too.

src/AutorestSwift/main.swift
```diff
@@ -31,7 +31,8 @@ guard let documentsUrl = FileManager.default.urls(for: .documentDirectory, in: .
     fatalError("Unabled to locate Documents directory.")
 }
 
-let sourceUrl = documentsUrl.appendingPathComponent("code-model-v4-2.yaml")
+let sourceUrl = documentsUrl.appendingPathComponent("communication-chat-code-model-v4-2.yaml")
+// let sourceUrl = documentsUrl.appendingPathComponent("code-model-v4-2.yaml")
```
tjprescott, line 35: I'd recommend reverting this change and just using the hard coded file `code-model-v4-2.yaml` for the time being.

templates/ChoiceSchema.stencil
```diff
@@ -0,0 +1,6 @@
+
+public enum {{ choice.language.default.name }}: {{ choice.choiceType.language.default.name }}, Codable {
```
tjprescott, line 2: This should just be `choice.name` and `choice.choiceType.name`. This ensures that it tries to pull from `language.swift` first and then `language.default` only as a backup.

templates/ChoiceSchema.stencil
```diff
@@ -0,0 +1,6 @@
...
+{% for aChoice in choice.choices %}
+    case {{ aChoice.value | lowercase }}
```
tjprescott, line 4: Remove the `| lowercase`. All naming and casing issues are the responsibility of the`SwiftNamer`. For now the casing will be messed up since it will pull from `language.default`

templates/ChoiceSchema.stencil
tjprescott, line 2: I think it will make more sense if the filename here reflects what the end-result Swift code is. So this would be `Enumeration.stencil` instead of `ChoiceSchema.stencil`. As we get into the more complex stuff, a single file will be composed of several schemas.

Open to ideas, but I think when the stencil describes the entire file, it should have a name that roughly matches the destination. For example, `Enumerations.swift` will be generated from the `Enumerations.stencil` and `StorageBlobClient.swift` will be generated from `ServiceClient.stencil`. 

For partial templates that are only used to compose others (like enum choices), I would recommend something like `EnumerationChoiceSnippet.stencil` and `HeaderSnippet.stencil`.  

Thoughts?

templates/Header.stencil
```diff
@@ -0,0 +1,9 @@
+// --------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
tjprescott, line 2: As stated above, I would recommend something in the name that clearly indicates that this isn't intended to be a standalone thing. `HeaderSnippet.stencil` or `HeaderPartial.stencil`.

templates/ObjectSchema.stencil
```diff
@@ -0,0 +1,20 @@
+// --------------------------------------------------------------------------
```
tjprescott, line 1: This should reference the Header stencil. Would also recommend a name like `Model.stencil` or `StructModel.stencil` (maybe we have `ClassModel.stencil` eventually?).

templates/ObjectSchema.stencil
```diff
@@ -0,0 +1,20 @@
...
+// Code generated by Microsoft (R) AutoRest Code Generator.
+// Changes may cause incorrect behavior and will be lost if the code is
+// regenerated.
+// --------------------------------------------------------------------------
+
+
+import AzureCore
+import Foundation
+
+public struct {{ object.language.default.name }} {
```
tjprescott, line 15: Same as above. `object.name`.

templates/ObjectSchema.stencil
```diff
@@ -0,0 +1,20 @@
...
+// license information.
+//
+// Code generated by Microsoft (R) AutoRest Code Generator.
+// Changes may cause incorrect behavior and will be lost if the code is
+// regenerated.
+// --------------------------------------------------------------------------
+
+
+import AzureCore
+import Foundation
```
tjprescott, line 13: Perhaps standard imports should be in a small stencil snippet as well?

templates/ObjectSchema.stencil
```diff
@@ -0,0 +1,20 @@
...
+
+public struct {{ object.language.default.name }} {
+    {% for property in object.properties %}
+    /* {{ property.language.default.description }} */
+    public let {{ property.language.default.name | lowercase }} : {{ String(property.schema.type) }}
```
tjprescott, line 18: Cut `| lowercase` and leave the casing to the namer. use `property.name` and `property.description`.

The comment needs to be a swift doc-comment (either /// or /** */) to be picked up by the Xcode help system and Jazzy doc generation.

If the description is nil, then no comment should be output. Looking at this I would think it would still output `/* */`

src/AutorestSwift/Code Generator/Manager.swift
sacheu: this is change by swiftlint

src/AutorestSwift/Code Generator/SwiftGenerator.swift
sacheu: Thanks. Removed.

Podfile
```diff
@@ -7,6 +7,7 @@ target 'AutorestSwift' do
 
   # Pods for AutorestSwift
   pod 'Yams'
+  pod 'Stencil', '~> 0.13.1'
```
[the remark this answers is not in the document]
sacheu: Yes. Thanks for pointing this out.

src/AutorestSwift/Models/Complex/ChoiceSchema.swift
sacheu: Great idea.

src/AutorestSwift/main.swift
sacheu: sorry. this is not meant for checkin

templates/Header.stencil
sacheu: Rename to HeaderSnippet.stencil

templates/ChoiceSchema.stencil
sacheu: Removed lowercase.

templates/ObjectSchema.stencil
sacheu: Move to IncludeSnippet.stencil

PR #71, opened 2020-07-24
Stencil Generation with View Models

tjprescott:
Uses a ViewModel approach for sending data to the Stencil templates, working around the issue that Stencil won't find computed Swift properties.


src/AutorestSwift/Code Generator/SwiftGenerator.swift
```diff
@@ -52,42 +49,28 @@ class SwiftGenerator: CodeGenerator {
...
-            do {
-                results.append(try stencilable.generateSnippet())
-            } catch {
-                print(error)
-            }
-        }
-    }
-
-    private func createModelFiles() throws {
+        let enumViewModel = EnumerationFileViewModel(from: model.schemas)
```
tjprescott, line 60: I will abstract this routine out to something more generic.

src/AutorestSwift/View Models/EnumerationFileViewModel.swift
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+struct EnumerationViewModel {
+    let name: String
+    let comment: String?
+    let type: String
+    let choices: [EnumerationChoiceViewModel]
+
+    init(from schema: ChoiceSchema) {
```
sacheu, line 47: Should we create a protocol say ChoiceSchemaProtocol and ChoiceSchema and SealConditionSchema implmenets and the init() just takes the ChoiceSchemaProtocol. this will save use from having duplicate code

tjprescott:
I added some more ViewModels and am assembling some best practices on writing ViewModels and Stencils that can be found here: https://github.com/Azure/autorest.swift/wiki/Guidelines-for-Writing-View-Models

PR #74, opened 2020-07-24
Add basic CI pipeline configuration

daviwil:
This change adds a basic CI pipeline configuration and a `package.json` that will later be used for publishing this package to GitHub/`npm` for use with AutoRest.  Here's an example of a successful CI run:

https://dev.azure.com/azure-sdk/public/_build/results?buildId=472695&view=logs&j=6614165d-8168-5766-9db6-f35f537f8e36

Once you're able to hook up the generator to AutoRest (RPC docs still underway) I'll set up a publishing pipeline for dev builds so that folks can try it out.


daviwil:
Switched the pipeline config to build from `master` now, should be good to go for all future PRs and commits to `master` 🎉

PR #76, opened 2020-07-27
Run SwiftFormat and SwiftLint in the generated code

sacheu:
Use .swiftformat and .swiftlint.yaml in the repo as the config files when running swiftFormat and swiftlint.


src/AutorestSwift/Code Generator/Manager.swift
```diff
@@ -141,4 +144,38 @@ class Manager {
             logger.log("Model file check: OK", level: .info)
         }
     }
+
+    private func runSwiftLintandFormat(atBaseUrl baseUrl: URL) throws {
```
tjprescott, line 148: I would just call this `formatCode(atBaseUrl:)`

sacheu:
> This will change the install directions right? Now the user would need swiftlint and swiftformat installed prior to running?

When developer use brew to install 'swiftlint', 'swiftformat', they will be installed to '/usr/local/bin/'

PR #77, opened 2020-07-28
Code generation aligns with swift package structure

sacheu:
* Create source code under 'Sources' directory
* Create a subdirectory used the Service Name under Sources directory which is the parent directory for all the code.
*.Create a Package.swift for the swift package
* Generated code can compile by go to `generated/sdk/Foo' and type 'swift build'.

### DIFF
--- a/src/AutorestSwift/Code Generator/SwiftGenerator.swift
+++ b/src/AutorestSwift/Code Generator/SwiftGenerator.swift
@@ -34,6 +34,8 @@ class SwiftGenerator: CodeGenerator {
 
     let baseUrl: URL
 
+    let targetName: String
+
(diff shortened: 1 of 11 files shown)


src/AutorestSwift/Code Generator/SwiftGenerator.swift
```diff
@@ -62,6 +64,7 @@ class SwiftGenerator: CodeGenerator {
             template: "EnumerationFile",
             toSubfolder: .models,
             withFilename: "Enumerations",
+            withTargetName: targetName,
```
tjprescott, line 67: I'm not clear why target name is propagated throughout all these methods. I'm not really sure what it is even supposed to be.

tjprescott:
LGTM. Please resolve the conflicts and then merge 👍

PR #78, opened 2020-07-28
Add unit tests and yaml files 

sacheu:
* Separate into executable and library: AutorestSwift and AutorestSwiftFramework
* add yaml created from testserver test swagger file
* add test caes to run the yaml files which passed


sacheu:
Close this PR. Will try without structure change.

PR #80, opened 2020-07-28
Lookup type of the Property from Schema and map that to Swift type

sacheu:
For each Property, look up its type from Schema and map that to corresponding swift type.

### DIFF
--- a/templates/ModelImportSnippet.stencil
+++ b/templates/ModelImportSnippet.stencil
@@ -1,0 +1,1 @@
+import Foundation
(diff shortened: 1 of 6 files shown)


templates/ModelImportSnippet.stencil
```diff
@@ -0,0 +1 @@
+import Foundation
```
tjprescott, line 1: What is the purpose of this snippet?

sacheu: The Model files does not need AzureCore as dependency. So I create a new snippet for model files only.

tjprescott: Hmm... okay. Maybe for the future, we can just pass a list of packages to our render command and they can be rendered with a for loop. Swift imports never get very complex. This is okay for now.

PR #81, opened 2020-07-28
Utility Script

tjprescott:
Generates SDK based on latest in-progress code gen and stages to the `azure-sdk-for-ios-pr` repo. Names to obfuscate work-in-progress service.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -297,5 +297,5 @@ package/
 src/obj/
 *.log
 IDEFindNavigatorScopes.plist
-.swiftpm/xcode/xcuserdata/travisprescott.xcuserdatad
-AutorestSwift.xcworkspace/xcuserdata/travisprescott.xcuserdatad
+.swiftpm/xcode/xcuserdata/
+AutorestSwift.xcworkspace/xcuserdata/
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -297,5 +297,5 @@ package/
 src/obj/
 *.log
 IDEFindNavigatorScopes.plist
-.swiftpm/xcode/xcuserdata/travisprescott.xcuserdatad
-AutorestSwift.xcworkspace/xcuserdata/travisprescott.xcuserdatad
+.swiftpm/xcode/xcuserdata/
+AutorestSwift.xcworkspace/xcuserdata/
```
tjprescott, line 301: This is to fix issues we've had with certain files sneaking into our commits.

Makefile
```diff
@@ -0,0 +1,38 @@
+UNAME = ${shell uname}
```
tjprescott, line 1: This is the workaround to copy the stencil files into the executable directory.

.gitignore
sacheu: Thanks for fixing this.

PR #86, opened 2020-07-29
Refactor view models

tjprescott:
This PR:
- breaks out several view models that were grouped into single files into their own files and adds doc comments to clarify their role.
- Adds more granular stencil templates to compose the Client.swift file
- Adds logic to generate the beginning of the ClientOptions.swift file
- Refactors all Code Model objects to be internal vice public.
- Adds methods to `CodeModel` and `Schemas` to search for a schema by name and type. Refactors KeyValue logic to use this method.


src/AutorestSwift/Models/Simple/Schemas.swift
```diff
@@ -27,86 +27,98 @@
...
-
-    public let any: [AnySchema]?
+    let unknowns: [Schema]?
+
+    let groups: [GroupSchema]?
+
+    let any: [AnySchema]?
+
+    /// Lookup a schema by name.
+    func schema(for name: String, withType type: AllSchemaTypes) -> Schema? {
```
tjprescott, line 114: @sacheu this is the logic that impacts `KeyValueViewModel`

PR #87, opened 2020-07-30
Separate QueryParameter and Http header into 2 groups: 

sacheu:
1. Can be set in the Initializer of Query Parameter and Http Header
2. Need to add nil check before adding to Query Parameter / Http Header

For the http header set in the Http Header initializer, it will be converted to use HttpHeaders enum key. (e.g. .accept)


src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -32,15 +32,32 @@ import Foundation
 struct KeyValueViewModel {
     let key: String
     let value: String
+    let valueNilable: Bool
```
tjprescott, line 35: We should use consistent naming. Elsewhere we've called this `required`. The right Swift term would be `optional`. I'm okay with either.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -32,15 +32,32 @@ import Foundation
...
         } else if let schema = model.schema(for: param.schema.name, withType: param.schema.type) {
             self.value = schema.name
+            self.valueNilable = false
         } else {
             self.value = ""
+            self.valueNilable = false
         }
     }
+
+    init(key: String, value: String) {
```
tjprescott, line 58: What is this initializer for, and why can you not set the optionality?

src/AutorestSwift/Models/Simple/Operation.swift
```diff
@@ -117,4 +117,9 @@ class Operation: Codable, LanguageShortcut {
         try container.encode(`protocol`, forKey: .protocol)
         if extensions != nil { try container.encode(extensions, forKey: .extensions) }
     }
+
+    /// Lookup a signatureParameter by name.
+    func signatureParameters(for name: String) -> Parameter? {
+        return signatureParameters?.first { $0.name == name }
```
tjprescott, line 123: The name of the method implies it will return a collection. `signatureParameter(for:)` communicates the intent better.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -32,15 +32,32 @@ import Foundation
...
     let key: String
     let value: String
+    let valueNilable: Bool
 
-    init(from param: Parameter, with model: CodeModel) {
-        self.key = param.serializedName!
-        if let constant = param.schema as? ConstantSchema {
-            self.value = "\"\(constant.value.value)\""
+    init(from param: Parameter, with model: CodeModel, and operation: Operation, using key: String? = nil) {
+        self.key = key ?? param.serializedName!
```
tjprescott, line 38: What is the case where we know the key to be able to just pass it in? Also, instead of using `!` it would make more sense to have a guard statement with a `fatalError(message)`. This will crash if `serializedName` is nil, but the fatalError will yield a better error message.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -32,15 +32,32 @@ import Foundation
...
+
+        if let constantSchema = param.schema as? ConstantSchema {
+            let isString: Bool = constantSchema.valueType.type == AllSchemaTypes.string
+            let val: String = constantSchema.value.value
+
+            self.value = isString ? "\"\(val)\"" : "\(val)"
+            self.valueNilable = false
```
tjprescott, line 45:
```suggestion
            let val = constantSchema.value.value
            if constantSchema.valueType.type == AllSchemaTypes.string {
              self.value = "\"\(val)\""
            } else {
              self.value = "\(val)" 
            }
            self.valueNilable = false
```

Fewer boolean variables should make it a little clearer.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -32,15 +32,32 @@ import Foundation
...
+        } else if let signatureParameter = operation.signatureParameters(for: param.name) {
+            self.value = param.name
```
tjprescott, line 47: Does this not need the same checking as other to determine whether it is a string or not?

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -35,6 +35,10 @@ struct OperationViewModel {
     let comment: ViewModelComment
     let params: [ParameterViewModel]
     let returnType: ReturnTypeViewModel?
+    // Query Params/Header can be placed inside QueryParameter Initializer
+    let queryInitParams: [KeyValueViewModel]?
+    let headerInitParams: [KeyValueViewModel]?
+    // Query Params/Header need to add Nil check
```
tjprescott, line 41: I don't think you need to separate these out. It is totally fine to instantiate an empty dictionary for queryParams and headers and then simply add each with subsequent lines (i.e. don't bother to populate in the initializer).

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -95,12 +110,40 @@ struct OperationViewModel {
...
 }
 
+private func isStandardHttpHeader(with headerName: String) -> HTTPHeaders? {
+    for header in HTTPHeaders.allCases where header.rawValue == headerName {
+        return header
+    }
+    return nil
+}
+
+private func convertHttpHeaderKey(from header: String?) -> String? {
```
tjprescott, line 142: I'm not clear on what the purpose of this is. Are you trying to map something like `x-ms-examples` to `examples`?  It would probably be better to use a string initializer on the enum that will spit out the actual enum value. This becomes brittle if we have to change the enum names.

src/AutorestSwift/View Models/PropertyViewModel.swift
```diff
@@ -40,31 +40,7 @@ struct PropertyViewModel {
         self.name = schema.name.toCamelCase
         self.comment = ViewModelComment(from: schema.description)
         let required = schema.required ?? false
-        self.type = getType(from: schema.schema, optional: !required)
+        self.type = schema.schema.swiftType + (!required ? "?" : "")
```
tjprescott, line 43: Yes, `swiftType` should operate like `getType` did, accepting an `optional: Bool` parameter to handle this in a centralized location.

templates/OperationSnippet.stencil
```diff
@@ -10,18 +10,42 @@ public func {{ op.name }}(
...
     var queryParams: [QueryParameter] = [
-        {% for param in op.queryParams %}
-        "{{ param.key }}" : {{ param.value}}
+        {% for param in op.queryInitParams %}
+        ("{{ param.key }}" , {{ param.value}})
         {% endfor %}
     ]
+    {% elif op.queryParams.count > 0 %}
+    var queryParams = [QueryParameter]()
+    {% endif %}
```
tjprescott, line 21: We want to avoid having conditional logic in our templates. I think it is totally fine to only the second version, and to simply add things after the fact.

templates/OperationSnippet.stencil
```diff
@@ -10,18 +10,42 @@ public func {{ op.name }}(
...
+
+    {% if op.queryParams.count > 0 %}
+    {% for param in op.queryParams %}
+    if {{ param.value }} != nil {
+        queryParams.append("{{ param.key }}", {{ param.value}})
+    }
+    {% endfor %}
+    {% endif %}
```
tjprescott, line 29: There's no need for this if statement if `op.queryParams` has a default value of empty collection.

templates/OperationSnippet.stencil
```diff
@@ -10,18 +10,42 @@ public func {{ op.name }}(
...
     var headers = HTTPHeaders([
-        {% for param in op.headerParams %}
-        "{{ param.key }}" : {{ param.value}}
+        {% for param in op.headerInitParams %}
+        {{ param.key }} : "{{ param.value}}",
         {% endfor %}
     ])
+    {% elif op.headerParams.count > 0 %}
+    var headers = HTTPHeaders()
+    {% endif %}
```
tjprescott, line 40: Same thing here--we want to avoid having if-else logic in our templates, if possible.

templates/OperationSnippet.stencil
```diff
@@ -10,18 +10,42 @@ public func {{ op.name }}(
...
+
+    {% if op.headerParams.count > 0 %}
+    {% for param in op.headerParams %}
+    if {{ param.value }} != nil {
+        headers["{{ param.key }}"] = {{ param.value}}
+    }
+    {% endfor %}
+    {% endif %}
```
tjprescott, line 48: Same here--no if needed if `op.headerParams` defaults to empty collection.

templates/OperationSnippet.stencil
```diff
@@ -10,18 +10,42 @@ public func {{ op.name }}(
     guard let url = self.url(forTemplate: urlTemplate) else { return }
 
     // Construct query
```
tjprescott, line 12: Since every operation will have a query params section and a header params section, I think it makes a lot of sense to move these sections into their own `stencil` files and add them via composition.

tjprescott:
> 2\. Need to add nil check before adding to Query Parameter / Http Header

Also, I would say that this is not strictly necessary. Since we are generating the code, it is probably fine to just use the raw header key strings. The enum is more useful when hand-authoring, but AzureCore will work with either form.

src/AutorestSwift/View Models/HttpHeaders.swift
```diff
@@ -0,0 +1,130 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+public enum HTTPHeaders: String, CaseIterable {
```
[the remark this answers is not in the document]
sacheu: How can we import from AzureCore directly? AzureCore has not published as a package/framework yet?

templates/OperationSnippet.stencil
sacheu: Make sense. Create separate stencil file for header and query param

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Renamed to optional.

src/AutorestSwift/Utils/String+Extensions.swift
```diff
@@ -94,4 +94,8 @@ extension String {
         let value = casedComponents.joined()
         return value
     }
+
+    func lowercaseFirstLetter() -> String {
+        return prefix(1).lowercased() + dropFirst()
```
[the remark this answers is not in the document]
sacheu: Method is removed.

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: removed.

templates/OperationSnippet.stencil
sacheu: Removed.

templates/OperationSnippet.stencil
sacheu: Removed.

src/AutorestSwift/View Models/HttpHeaders.swift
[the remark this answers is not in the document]
tjprescott: Well, we can't right now, but ultimately we would just import AzureCore and reference this. It's fine to keep this in for now, but we would need a TODO linking to a tracking issue in Github.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -32,15 +32,43 @@ import Foundation
...
-            self.value = "\"\(constant.value.value)\""
-        } else if let schema = model.schema(for: param.schema.name, withType: param.schema.type) {
-            self.value = schema.name
+    init(
+        from parameter: Parameter? = nil,
+        with codemodel: CodeModel? = nil,
+        and operation: Operation? = nil,
+        using key: String? = nil,
+        modelValue: String? = nil
+    ) {
```
tjprescott, line 44: It seems like there's a lot of complexity in this initializer but I don't understand the reason for it. Perhaps we could break this into multiple initializers:

```swift 
init(from constant: ConstantSchema) {
let isString: Bool = constantSchema.valueType.type == AllSchemaTypes.string
  let val: String = constantSchema.value.value
  self.value = isString ? "\"\(val)\"" : "\(val)"
  self.optional = false
}

init(value: String, optional: Bool) {
  ...
}

init(from parameter: parameter, with model: CodeModel, and operation: Operation) {
  ...
}
```

Though I would still like to see a doc comment that explains where each initializer would be used.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -32,15 +32,43 @@ import Foundation
 struct KeyValueViewModel {
     let key: String
     let value: String
+    // Flag indicates if value is optional
+    let optional: Bool
```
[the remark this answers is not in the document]
sacheu: Rename to `optional`

src/AutorestSwift/Models/Complex/Schema.swift
```diff
@@ -64,4 +64,28 @@ class Schema: Codable, LanguageShortcut {
...
+                swiftType = "[\(arraySchema.elementType.name)]"
+            } else {
+                swiftType = "[\(name)]"
+            }
+        case AllSchemaTypes.dateTime:
+            swiftType = "Date"
+        case AllSchemaTypes.integer:
+            swiftType = "Int"
+        default:
+            swiftType = name
```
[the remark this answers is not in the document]
sacheu: Change default as fatalError.

src/AutorestSwift/Models/Complex/Schema.swift
```diff
@@ -64,4 +64,25 @@ class Schema: Codable, LanguageShortcut {
 
     /// Additional metadata extensions dictionary
     let extensions: AnyCodable?
+
+    var swiftType: String {
```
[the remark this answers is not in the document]
sacheu: Changed.

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Bring down into 2 initializers.
1. Add logic to look up if the param is a Constant or a SignatureSignureParameter.  
if it is constant, the value will the constant value
If it is from Signature Parameter, the value will be name of the Signature Parameter.

2.Add a initializer just takes the key, value pair as Strings.
This is for building http header for contentType and Accept.  The values of them are taking from knownMediaType and MediaTypes of request and response.

PR #88, opened 2020-07-30
Decompose Operation.Stencil

tjprescott:
Creates smaller stencil files for:

- Handling options
- Building the Request
- Handling the response, which has three variants: Pageable, Decoding or StatusOnly.


templates/ClientMethodOptionsFile.stencil
```diff
@@ -0,0 +1,7 @@
+{% include "HeaderSnippet.stencil" %}
+
+{% include "ImportSnippet.stencil" %}
+
+{% for model in models %}
+    {% include "ClientMethodOptionsSnippet.stencil" model %}
```
sacheu, line 6: Cool. I did not know you can pass in parameter when include another stencil.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
```diff
@@ -26,13 +26,22 @@
 
 import Foundation
 
+enum ReturnTypeEnum: String {
+    case pageable
+    case decoding
```
sacheu, line 31: The name `decoding` does not seem to consistent with `pageable` , `statuscode`.  `pageable` , `statuscode` matches the name of Enum 'ReturnType', but decoding seems to say what you are doing to do with the return type.

tjprescott: I agree, I don't love these names. I could add doc comments to clarify or we could just choose different ones. Thoughts?

```Swift
/// returns PagedCollection<Codable>
case pageable
/// returns some Codable type
case decoding
/// no return data. Simply checks the status code
case statusCode
```

tjprescott: So we aren't really trying to match the code model's ReturnType enum here. Most of those things would fall into the "decoding" bucket where we can just decode them with `JSONDecoder.decode(Codable.type, data: data)`. It's more trying to reflect the flavor of the code that needs to go in that section to accomplish the goal.

templates/ClientMethodOptionsFile.stencil
tjprescott: Yes--quite a useful feature! It means we can get away with fewer ViewModel structs 😁 
https://stencil.fuller.li/en/latest/builtins.html#include

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
sacheu: maybe 'ResponseWithBody' ?

templates/ClientMethodOptionsSnippet.stencil
```diff
@@ -0,0 +1,23 @@
...
+{% endfor %}
+
+    /// Initialize a `{{ model.name }}` structure.
+    /// - Parameters:
+{% for property in model.properties %}
+    ///   - {{ property.name }}: {{ property.comment }}
+{% endfor %}
+    public init(
+    {% for property in model.properties %}
+        {{ property.name }}: {{ property.type }} = nil,
```
sacheu, line 16: do we need to check if the type is an optional type before setting it to "nil' as default value?

templates/ClientMethodOptionsSnippet.stencil
```diff
@@ -0,0 +1,23 @@
+/// User-configurable options for the `{{ model.clientName }}.{{ model.operationName }}` operation.
+public struct {{ model.name }}: AzureOptions {
+
+{% for property in model.properties %}
+    {{ property.comment }}
+    public let {{ property.name }}: {{ property.type }}{{property.defaultValue}}
```
sacheu, line 6: We need a "=" between the type and defaultValue {{ property.type }}{{property.defaultValue}}

templates/ClientMethodOptionsSnippet.stencil
tjprescott: Possibly. A "required option" seems like an oxymoron, but I could see scenarios where it might make sense. The entire options object is optional, which implies that even if there was a "required option", if you aren't required to provide the options, then it doesn't much matter. That's why the defaults are all nil in the options struct.

tjprescott: I'd say let's leave it for now and see how that plays out. I suspect that if there is a "required option" then the service will return an error. I will also ask some autorest folks about this scenario.

templates/ClientMethodOptionsSnippet.stencil
tjprescott: The `defaultValue` primitive is either "" or " = {value}"

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
tjprescott: I would be okay with that. I pushed one option, but I might like this better:
```
case pagedBody
case body
case noBody
```
Thoughts?

sacheu: Sound good!

tjprescott: Or
```swift
enum BodyType: String {
case paged
case normal
case none
}
```

sacheu: i like better
```
case pagedBody
case body
case noBody
```

tjprescott: Updated. Let me know if it looks good to you.

sacheu: Looks good.

sacheu:
PR looks good. Approved.

PR #98, opened 2020-07-31
Fix client code generation to be enabled generated code to compile cleanly

### LINKED ISSUE #90: Update logic to use appropriate OperationResponseSnippet
#88 adds stencils for different OperationResponseSnippets, but the logic in `ReturnTypeViewModel` does not actually set the appropriate type (it defaults to decoding style).

- [ ] Update `ReturnTypeViewModel` to set the appropriate type for an operation. Once the appropriate type is set, the correct snippet should automatically be generated.

### LINKED ISSUE #93: Populate API Versions enum in ServiceClient.swift
Example from Storage:

```Swift
    /// API version of the Azure Storage Blob service to invoke. Defaults to the latest.
    public enum ApiVersion: String {
        /// API version "2019-02-02"
        case v20190202 = "2019-02-02"

        /// The most recent API version of the Azure Storage Blob service
        public static var latest: ApiVersion {
            return .v20190202
        }
    }
```

### LINKED ISSUE #94: Populate PipelineContext objects
We must populate the PipelineContext with appropriate data. 
Here are the currently known ContextKeys:
```swift
public enum ContextKey: String {
    case allowedStatusCodes
    case allowedHeaders
    case cancellationToken
    case deserializedData
    case requestStartTime
    case xmlMap
}
```

For our current work, we should worry about:
- allowedStatusCodes: should probably be added to every request. Don't count on the default.
- cancellationToken: should be added to every request, if present.

xmlMap is important, but not for our current scope since the service doesn't use XML (hooray). 

allowedHeader, deserializedData, and requestStartTime are all used internally by AzureCore, so we shouldn't have to do anything with them.

sacheu:
Fix #90 
Fix #94  - Add allowed status code in pipeline context
Fix #93 
Fixes other issues to enable generated client code to compile cleanly

### DIFF
--- a/src/AutorestSwift/View Models/KeyValueViewModel.swift
+++ b/src/AutorestSwift/View Models/KeyValueViewModel.swift
@@ -30,9 +30,12 @@ import Foundation
 /// Example:
 ///     "key" = value
 struct KeyValueViewModel {
+    /// key of the Key-Value pair
     let key: String
-    let paramName: String?
+    /// value of the Key-Value pair
     let value: String
+    /// nane of the parameter where the 'value' is retrieved from.
+    let paramName: String?
     // Flag indicates if value is optional
     let optional: Bool
 
--- a/src/AutorestSwift/View Models/KeyValueViewModel.swift
+++ b/src/AutorestSwift/View Models/KeyValueViewModel.swift
@@ -63,8 +63,8 @@ struct KeyValueViewModel {
             if swiftType.starts(with: "String") {
                 self.value = param.name
             } else {
-                // Convert the Swift type into String
-                self.value = "String(\(param.name))"
+                // Convert into String in generated code
+                self.value = "String(\(param.name)!)"
             }
         } else {
             self.value = ""
--- a/src/AutorestSwift/View Models/KeyValueViewModel.swift
+++ b/src/AutorestSwift/View Models/KeyValueViewModel.swift
@@ -31,6 +31,7 @@ import Foundation
 ///     "key" = value
 struct KeyValueViewModel {
     let key: String
+    let paramName: String?
     let value: String
     // Flag indicates if value is optional
     let optional: Bool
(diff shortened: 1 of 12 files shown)


templates/OperationRequestSnippet.stencil
```diff
@@ -3,6 +3,9 @@ guard let requestUrl = url.appendingQueryParameters(queryParams) else { return }
 guard let request = try? HTTPRequest(method: .{{ op.method }}, url: requestUrl, headers: headers) else { return }
 let context = PipelineContext.of(keyValues: [
    // TODO: Context keys
```
tjprescott, line 5: You can remove the TODO :grin:

templates/ClientOptionsFile.stencil
```diff
@@ -28,5 +30,8 @@ public struct {{ model.name }}Options: AzureClientOptions {
         self.logger = logger
         self.telemetryOptions = telemetryOptions
         self.transportOptions = transportOptions ?? TransportOptions()
+
+        // TODO: To be implemented
+        self.cancellationToken = nil
```
tjprescott, line 35: Is this a follow-up item, or do you plan to put this into this PR? The AzureCore PR has been merged.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
```diff
@@ -42,9 +42,8 @@ struct ReturnTypeViewModel {
     let name: String
     let strategy: String
 
-    init(from objectType: String) {
-        self.name = objectType
-        // TODO: Logic here to decide what type this should be
-        self.strategy = BodyType.body.rawValue
+    init(from response: ResponseViewModel?) {
+        self.name = response?.objectType ?? "Void"
+        self.strategy = response?.objectType != nil ? BodyType.body.rawValue : BodyType.noBody.rawValue
```
tjprescott, line 47: What about for the paged responses?

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -31,6 +31,7 @@ import Foundation
 ///     "key" = value
 struct KeyValueViewModel {
     let key: String
+    let paramName: String?
```
tjprescott, line 34: Could you update the example in the doc string? It would make it clearer without having to look at the stencil file what it's role is in the code gen.

templates/OperationRequestSnippet.stencil
sacheu: Done. Thanks.

templates/ClientOptionsFile.stencil
sacheu: a Follow up item. Not part of this pr.

sacheu: Fixed.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
sacheu: added handling for paged response

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Added doc string

PR #99, opened 2020-07-31
Add support for encoding request body

### LINKED ISSUE #96: Generate code to populate the request body
We have not done this yet in AzureStorageBlob, so we need to investigate how to generate this code.

sacheu:
Fix #96

### DIFF
--- a/templates/OperationRequestBodySnippet.stencil
+++ b/templates/OperationRequestBodySnippet.stencil
@@ -1,6 +1,10 @@
 // Construct request
-let requestBodyjsonData = try! JSONEncoder().encode( {{ op.request.objectName }})
-let requestBodyData: Data? =  String(data: requestBodyjsonData, encoding: .utf8)!.data(using: .utf8)
+guard let requestBody = try? JSONEncoder().encode( {{ op.request.objectName }} ) else {
+    throw AzureError.sdk("Failed to encode request body as json.")
+}
 
-guard let requestUrl = url.appendingQueryParameters(queryParams) else { return }
-guard let request = try? HTTPRequest(method: .{{ op.request.method }}, url: requestUrl, headers: headers, data: requestBodyData) else { return }
+guard let requestUrl = url.appendingQueryParameters(queryParams) else {
+    throw AzureError.sdk("Failed to append query parameters to url")
+}
+
+guard let request = try? HTTPRequest(method: .{{ op.request.method }}, url: requestUrl, headers: headers, data: requestBody) else { throw AzureError.sdk("Failed to construct Http request") }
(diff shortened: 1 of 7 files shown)


templates/OperationRequestBodySnippet.stencil
```diff
@@ -0,0 +1,6 @@
+let requestBodyjsonData = try! JSONEncoder().encode( {{ op.request.objectName }})
```
tjprescott, line 2: We shouldn't put a `!` into production code. If the larger method throw we can just let this throw. If not we can just wrap this in a `do-catch` and log the error.

templates/OperationRequestBodySnippet.stencil
```diff
@@ -0,0 +1,6 @@
...
+let requestBodyData: Data? =  String(data: requestBodyjsonData, encoding: .utf8)!.data(using: .utf8)
```
tjprescott, line 3: Why the switch from Data -> String -> Data? Suggest removing or adding a comment to clarify

tjprescott: Same issue as above. Don't use `!` in production code because it's a hard crash with "unwrapped nil value". Also, I feel like if it should error out when it can't encode the object, then it should also error out if the conversion to and from String actually yields nil. In other words, I feel like it should look something like:

```Swift
guard let requestBodyData = String(...).data(using: .utf8) else {
   // some error here
}
```

Again, not that you don't need to specify the data type on the left of the equals since it can be inferred (though it did help me spot this comment so maybe that's not a bad idea!).

templates/OperationSendRequestSnippet.stencil
```diff
@@ -0,0 +1,20 @@
+{% if op.request.hasBody %}
+{% include "OperationRequestBodySnippet.stencil" op %}
+{% else %}
+{% include "OperationRequestSnippet.stencil" op %}
+{% endif %}
+
+// Send reqeust
+let context = PipelineContext.of(keyValues: [
+   // TODO: Context keys
```
tjprescott, line 9: This TODO should be removed.

tjprescott: Looks like this was missed?  This TODO isn't needed in the stencil anymore.

templates/OperationSnippet.stencil
```diff
@@ -5,13 +5,13 @@ public func {{ op.name }}(
     {% endfor %}
     withOptions options: {{ model.name }}Options? = nil,
     completionHandler: @escaping HTTPResultHandler<{{ op.returnType.name }}>
-) {
+) throws {
```
tjprescott, line 8: Hmm... if the method body didn't already throw, we shouldn't make it throw. Instead of throwing AzureErrors, we should probably log an error and early return. We had UX study feedback that the more things that are throwable the more confusing it is for customers, so this is actually a big usability change.

PR #100, opened 2020-07-31
Add Service client doc comments

### LINKED ISSUE #95: Service client doc comments should be full-featured
Example of current generation:
```swift
    /// Posts a read receipt event to a thread, on behalf of a user.
    public func sendReadReceipt(
        threadId _: String,

        correlationVector: String?,

        body _: PostReadReceiptRequest,

        completionHandler: @escaping HTTPResultHandler<>
    )
```

However, this should be a fully-featured doc comment since it is used by XCode's code hinting and as the basis for documentation using Jazzy. It should look something like:
```swift
    /// Posts a read receipt event to a thread, on behalf of a user.
    /// - Parameters:
    ///   - threadId: The `String` thread id
    ///   - correlationVector: An optional `String` to send.
    ///   - body: The `ReadReceiptRequest` to send.
    ///   - completionHandler: A completion handler that receives a status code on
    ///     success.
    public func sendReadReceipt(
        threadId _: String,

        correlationVector: String?,

        body _: PostReadReceiptRequest,

        completionHandler: @escaping HTTPResultHandler<>
    )
```

sacheu:
Fix #95

### DIFF
--- a/templates/OperationSnippet.stencil
+++ b/templates/OperationSnippet.stencil
@@ -1,10 +1,8 @@
 {{ op.comment }}
 /// - Parameters:
-{% for param in op.params %}
-///   - {{ param.name }}: {{ param.comment }}
-{% endfor %}
-///   - options: A list of options for the operation
-///   - completionHandler: A completion handler that receives a status code on
+{{ op.paramscomment }}
+///    - options: A list of options for the operation
+///    - completionHandler: A completion handler that receives a status code on
 ///     success.
 public func {{ op.name }}(
     {% for param in op.params %}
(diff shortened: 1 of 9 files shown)


templates/OperationSnippet.stencil
```diff
@@ -1,4 +1,11 @@
 {{ op.comment }}
+/// - Parameters:
+{% for param in op.params %}
+///   - {{ param.name }}: {{ param.comment }}
+{% endfor %}
+///   - options: A list of options for the operation
+///   - completionHandler: A completion handler that receives a status code on
+///     success.
```
tjprescott, line 8: The only problem here is that we will have lines generated even where there is no doc-comment. It might make sense to build the entire comment up in `ViewModelComment`.

sacheu: Good point.

Issue #89, opened 2020-07-30
Generate OperationOptions objects and connect them to their operation signatures


tjprescott:
#88 adds stencil files with which to generate the per-operation Options structs, but it does not actually generate them. Work remaining:

- [x] Generate the `ClientMethodOptions.swift` file in the Options folder as per AzureStorageBlob. It will contain all of the structs, just like the `Enumerations.swift` file. (PR #102)
- [x] Modify the `OperationSnippet.stencil` file to reference the options object  and only refer to URL and body encoded parameters in the signature. (PR #102)
- [x] Modify the `OperationOptionsSnippet.stencil` to put each option in the appropriate location (query params or header). (#111)
- [x] Update the method signature comment to remove items that were relocated to the options object. (#111)

sacheu:
Fix by #102

PR #102, opened 2020-08-03
Adding Options struct for each operation

sacheu:
Code generate 'options' struct for each operation optional parameters.
* For optional schema signature parameter and request signature parameter, move them to properties in a 'options' for each operation
* Change the generated code to retrieve value of the optional parameters from the options struct
* Change the generated code to get logger from AzureCommunicationChatClientOptions
* Change the generated code to get cancellation token from  'options' struct
* Small refactoring in OperationViewModel: Rename variable


sacheu:
All generated code still compiles with this PR.

sacheu:
Fixes #89

AutorestSwift.xcodeproj/xcuserdata/sacheu.xcuserdatad/xcschemes/xcschememanagement.plist
```diff
@@ -12,12 +12,12 @@
 		<key>AutorestSwiftPackageDescription.xcscheme_^#shared#^_</key>
```
tjprescott, line 12: This file shouldn't be under source control. I thought we added `xcuserdata/*` to the `.gitignore`?

src/AutorestSwift/View Models/ClientMethodOptionsViewModel.swift
```diff
@@ -37,16 +37,17 @@ struct ClientMethodOptionsViewModel {
...
 
-    let properties: [PropertyViewModel]
+    // The list of properites for the options object
+    let properties: [ParameterViewModel]
 
-    init(from operation: Operation, with model: CodeModel) {
+    /// The variable of the options object in the generated code
+    let variableName: String
+
+    init(from operation: Operation, with model: CodeModel, properties: [ParameterViewModel]) {
```
tjprescott, line 46: ViewModel initializers shouldn't take other view models as parameters. They should take the underlying model objects.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -108,8 +102,26 @@ struct OperationViewModel {
...
+        // All the optional parameters will be in optionalParams
+        var signatureParams = [ParameterViewModel]()
+        var optionalParams = [ParameterViewModel]()
+        for param in operation.signatureParameters ?? [] {
+            if param.required ?? false {
+                signatureParams.append(ParameterViewModel(from: param))
+            } else {
+                optionalParams.append(ParameterViewModel(from: param))
+            }
+        }
```
tjprescott, line 117: So the logic here isn't really whether the parameter is required or not. It's anything that isn't path or body encoded (i.e. query or header params), which you could get with something like 

```Swift
let optionsParams = operation.signatureParameters.filter { param in
  guard let location = param.protocol.http?.in else { return false }
  let allowed = [.header, .query]
  return allowed.contains(location)
}
```

However, I still think it would be good to have an assert to validate our assumption that anything in options should in fact be optional.

```Swift
for param in optionsParams {
  assert(param.required == false, "Unexpectedly found a required 'option'... \(param.name)")
}
```

tjprescott: Likewise, signature params *should* be anything path encoded first, followed by a body parameter, if there is one.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -145,6 +157,11 @@ struct OperationViewModel {
         self.requiredQueryParams = requiredQueryParams
         self.requiredHeaders = requiredHeaders
         self.uriParams = uriParams
+        self.clientMethodOptionsViewModel = ClientMethodOptionsViewModel(
+            from: operation,
+            with: model,
+            properties: optionalParams
```
tjprescott, line 163: So this should just pass in the optionalParams as `[Parameter]` and let `ClientMethodOptionsViewModel` create the view models from that. If at any point the ClientMethodOptionsViewModel might want more info about a parameter, it would likely not be in a ParameterViewModel. That's why it's best to just pass it all of the info and only strip the information down at the very end.

src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -58,8 +58,15 @@ struct RequestViewModel {
...
-        self.objectName = request.signatureParameters?.first?.name
+        // TODO: only support the first signature parameter in Reqest now
+        let firstSignatreParam = request.signatureParameters?.first
+        self.objectType = firstSignatreParam?.schema.name
+        if firstSignatreParam?.required ?? false {
+            self.objectName = firstSignatreParam?.name
+        } else {
+            let optionName = "\(operation.name)Options?".toCamelCase
+            self.objectName = "\(optionName).\(firstSignatreParam?.name ?? "")"
+        }
```
tjprescott, line 69: There's a typo here (signatre => signature) but I'm not understanding the comment. What is special about the first signature parameter? Why would the options object ever not have the same name as the method?

templates/OperationHeaderSnippet.stencil
```diff
@@ -5,7 +5,7 @@ headers["{{ header.key }}"] = {{ header.value }}
 {% endfor %}
 
 {% for header in op.optionalHeaders %}
-if {{ header.paramName }} != nil {
+if let {{ header.paramName }} = {{ op.clientMethodOptionsViewModel.variableName }}?.{{ header.paramName }} {
```
tjprescott, line 8: Since `op` is already a view model, I think we can shorten the `clientMethodOptionsViewModel` name to something like `options`. :smile:

I don't think we've append "viewModel" to any other variable names in other view models. 
```Swift
    // Example from OperationViewModel
    let optionalQueryParams: [KeyValueViewModel]?
    let optionalHeaders: [KeyValueViewModel]?
    let pipelineContext: [KeyValueViewModel]?
```

So I think we can follow that precedent.

templates/OperationOptionsSnippet.stencil
```diff
@@ -1,5 +1,5 @@
 // Process endpoint options
-if let options = options {
+if let {{ op.clientMethodOptionsViewModel.variableName }} = {{ op.clientMethodOptionsViewModel.variableName }} {
```
tjprescott, line 2: The variable name for the options in the method signature should always be `options`, so I'm not sure why the need to dynamically generate?

templates/OperationQueryParamSnippet.stencil
```diff
@@ -6,7 +6,7 @@ var queryParams: [QueryParameter] = [
 ]
 
 {% for param in op.optionalQueryParams %}
-if {{ param.paramName }} != nil {
+if let {{ param.paramName }} = {{ op.clientMethodOptionsViewModel.variableName }}?.{{ param.paramName }} {
```
tjprescott, line 9: We should be consistent here and not use both `variableName` and `paramName` when they are logically the same thing. We should change one or the other. I suspect paramName is already used more extensively throughout, but I would be okay with using variableName as long as we update it in all places.

templates/OperationRequestBodySnippet.stencil
```diff
@@ -1,15 +1,15 @@
 // Construct request
 guard let requestBody = try? JSONEncoder().encode( {{ op.request.objectName }} ) else {
-    options?.logger.error("Failed to encode request body as json.")
+    self.options.logger.error("Failed to encode request body as json.")
```
tjprescott, line 3: Does swiftlint strip out these `self.` usages in the final code? if so, we can remove them here as well for easier maintainability.

templates/OperationSendRequestSnippet.stencil
```diff
@@ -10,7 +10,7 @@ let context = PipelineContext.of(keyValues: [
    {{ param.key }} : {{ param.value}} as AnyObject
    {% endfor %}
 ])
-if let cancellationToken = options?.cancellationToken {
+if let cancellationToken = {{ op.clientMethodOptionsViewModel.variableName }}?.cancellationToken {
```
tjprescott, line 13: Again, this variable should always be called just `options` (i.e. not `listThreadOptions` or anything like that) so I'm not seeing the need for dynamically generated names here.

templates/OperationSnippet.stencil
```diff
@@ -5,10 +5,10 @@
 ///    - completionHandler: A completion handler that receives a status code on
 ///     success.
 public func {{ op.name }}(
-    {% for param in op.params %}
+    {% for param in op.signatureParams %}
     {{ param.name }}: {{ param.type }}{{ param.defaultValue }},
     {% endfor %}
-    withOptions options: {{ model.name }}Options? = nil,
+    withOptions {{ op.clientMethodOptionsViewModel.variableName }}: {{ op.clientMethodOptionsViewModel.name }}? = nil,
```
tjprescott, line 11: Especially since this is an internal label, there's really no reason NOT to use a static name. It simplifies the stencil files greatly.

templates/OperationUrlSnippet.stencil
```diff
@@ -6,6 +6,6 @@ let pathParams = [
     {% endfor %}
 ]
 guard let url = self.url(forTemplate: urlTemplate, withKwargs: pathParams) else {
-    options?.logger.error("Failed to construct url")
+    self.options.logger.error("Failed to construct url")
```
tjprescott, line 9: Same question about `self.` as above. If it doesn't survive the swiftlint/swiftformat combo, we shouldn't have it in the stencil.

AutorestSwift.xcodeproj/xcuserdata/sacheu.xcuserdatad/xcschemes/xcschememanagement.plist
sacheu: Yeah. I added 'AutorestSwift.xcodeproj/xcuserdata/'. to .gitigore. Not sure why git does not ignore this file.

templates/OperationQueryParamSnippet.stencil
sacheu: Rename to paramName.

src/AutorestSwift/View Models/ClientMethodOptionsViewModel.swift
sacheu: Fixed.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Changed per feedback.

templates/OperationRequestBodySnippet.stencil
sacheu: We need to add 'self.options' to reference the property 'options' of the client class.

templates/OperationUrlSnippet.stencil
sacheu: We need to add 'self.' to reference 'options' property of the client class.

AutorestSwift.xcodeproj/xcuserdata/sacheu.xcuserdatad/xcschemes/xcschememanagement.plist
sacheu: Revert the change.

templates/OperationRequestBodySnippet.stencil
sacheu: We need to add 'self' to reference the `options` property of the generated client class.

templates/OperationSendRequestSnippet.stencil
sacheu: Changed per feedback.

templates/OperationSnippet.stencil
sacheu: Changed per feedback.

templates/OperationUrlSnippet.stencil
sacheu: We need to reference 'options' property of the client class.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Changed per feedback.

src/AutorestSwift/View Models/RequestViewModel.swift
sacheu: We only support 1  request object (the class which we json encode to build the http payload) now  and the signure parameter is used to built the objectName of the request object ` self.objectName = firstSignatreParam?.name` . Currently 'self.objectName' is a String (not an array), that is why we only take the first signature parameters for the objectName.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -91,11 +85,23 @@ struct OperationViewModel {
             }
         }
 
+        var signatureParams = filterParams(for: operation.signatureParameters ?? [], with: [.path, .uri])
+        var optionsParams = filterParams(for: operation.signatureParameters ?? [], with: [.header, .query, .body])
```
tjprescott, line 89: You could also make `filterParams(for: [Parameter]?, ...)` then you can handle the nil case with a guard and not need the `?? []`.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -145,7 +153,24 @@ struct OperationViewModel {
         self.requiredQueryParams = requiredQueryParams
         self.requiredHeaders = requiredHeaders
         self.uriParams = uriParams
+        self.clientMethodOptionsViewModel = ClientMethodOptionsViewModel(
```
tjprescott, line 156: We shouldn't add `viewModel` to the property name. It just makes the the code verbose and we don't do that anywhere else that we have a view model property inside another view model.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -145,7 +153,24 @@ struct OperationViewModel {
...
+    }
+}
+
+private func filterParams(for params: [Parameter], with allowed: [ParameterLocation]) -> [Parameter] {
+    let optionsParams = params.filter { param in
+        guard let httpParam = param.protocol.http as? HttpParameter else { return false }
+        return allowed.contains(httpParam.in)
     }
+    return optionsParams
 }
```
tjprescott, line 174:
```suggestion
private func filterParams(for params: [Parameter]?, with allowed: [ParameterLocation]) -> [Parameter] {
    guard params != nil else { return [Parameter]() }
    let optionsParams = params.filter { param in
        guard let httpParam = param.protocol.http as? HttpParameter else { return false }
        return allowed.contains(httpParam.in)
    }
    return optionsParams
}
```

Will let you simplify the invocations by not needed `?? []`

src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -50,16 +49,24 @@ struct RequestViewModel {
...
-        for param in request.signatureParameters ?? [] {
-            params.append(ParameterViewModel(from: param))
+        // TODO: only support the first signature parameter in Reqest now
+        let firstSignatureParam = request.signatureParameters?.first
+        self.objectType = firstSignatureParam?.schema.name
+        if isRequired(param: firstSignatureParam) {
+            self.objectName = firstSignatureParam?.name
+        } else {
+            self.objectName = "options?.\(firstSignatureParam?.name ?? "")"
         }
```
tjprescott, line 59: I guess I still don't understand this logic. What scenarios is it trying to describe?

src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -50,16 +49,24 @@ struct RequestViewModel {
...
 }
+
+private func isRequired(param: Parameter?) -> Bool {
+    guard let httpParam = param?.protocol.http as? HttpParameter else { return false }
+
+    if httpParam.in == .path || httpParam.in == .uri {
+        return true
+    } else {
+        return false
+    }
```
tjprescott, line 71:
```suggestion
    let required = [.path, .uri]
    return required.contains(httpParam.in)
```

src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -50,16 +49,24 @@ struct RequestViewModel {
...
         }
-        self.params = params
-
-        self.objectType = request.signatureParameters?.first?.schema.name
-        self.objectName = request.signatureParameters?.first?.name
         self.hasBody = objectType != nil
     }
 }
+
+private func isRequired(param: Parameter?) -> Bool {
```
tjprescott, line 64: I'm not sure "isRequired" is the right phrase. In rare cases, you can have URI paramters that are optional. Maybe `belongsInSignature(...)`?

templates/OperationSnippet.stencil
```diff
@@ -5,10 +5,10 @@
 ///    - completionHandler: A completion handler that receives a status code on
 ///     success.
 public func {{ op.name }}(
-    {% for param in op.params %}
+    {% for param in op.signatureParams %}
     {{ param.name }}: {{ param.type }}{{ param.defaultValue }},
     {% endfor %}
-    withOptions options: {{ model.name }}Options? = nil,
+    withOptions options: {{ op.clientMethodOptionsViewModel.name }}? = nil,
```
tjprescott, line 11: Yes, the property should just be `options` or `methodOptions`.

src/AutorestSwift/View Models/RequestViewModel.swift
tjprescott: So we are looking for the body parameter? I think it would make sense to iterate through the signatureParameters and filter by those that have `http.in = .body`. We should then assert there is only 1 and continue. Instead of calling it `firstSignatureParam` I would call it "bodyParam" and put a comment that we are looking for a body-encoded parameter, if any.

templates/OperationHeaderSnippet.stencil
tjprescott: I think this one was missed.

src/AutorestSwift/View Models/RequestViewModel.swift
tjprescott: I didn't see your earlier reply, so I do understand this a little better now. I think we should clarify in the comment that we are looking for a body-encoded parameter, and then we should filter the signatureParams by whether they have `http.in == .body`. Since we only expect one, we should assert that there's only one and then continue with our assumption. This ensures the code will crash if the assumption is invalidated but only when i development mode.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Good point.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Good catch. Renamed.

src/AutorestSwift/View Models/RequestViewModel.swift
sacheu: Renamed.

src/AutorestSwift/View Models/RequestViewModel.swift
sacheu: Appy suggestion.

templates/OperationSnippet.stencil
sacheu: Fixed.

PR #105, opened 2020-08-04
Comment out build step from git_sync.sh

tjprescott:
The build steps always fail anyways, so the only thing this script does it copy the generated code from running the program in XCode to the azure-sdk-for-ios-pr repo.


sacheu:
You change looks good.

PR #107, opened 2020-08-04
Fix required in value schema

sacheu:


src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -160,6 +160,7 @@ struct OperationViewModel {
         )
 
         for param in optionsParams {
+            // let required = param.required ?? false
```
tjprescott, line 163: You can delete this.

src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -78,6 +78,6 @@ private func getBodyParameters(signatureParameters: [Parameter]?) -> [Parameter]
 private func belongsInSignature(param: Parameter?) -> Bool {
     guard let httpParam = param?.protocol.http as? HttpParameter else { return false }
 
-    let required = [ParameterLocation.path, ParameterLocation.uri]
+    let required = [ParameterLocation.path, ParameterLocation.uri, ParameterLocation.body]
```
tjprescott, line 81:
```suggestion
    let required: [ParameterLocation] = [.path, .uri, .body]
```

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Removed.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -61,7 +61,7 @@ struct KeyValueViewModel {
         } else if let signatureParameter = operation.signatureParameter(for: param.name) {
             // value is referring a signautre parameter, no need to wrap as String
             self.paramName = param.name
-            self.optional = signatureParameter.required ?? true
+            self.optional = !signatureParameter.required
```
tjprescott, line 64: What is the `!` for here? I've not seen that syntax before, but it seems like a "crash if nil" thing that we should avoid.

src/AutorestSwift/View Models/RequestViewModel.swift
sacheu: Changed.

PR #111, opened 2020-08-05
Fix bugs when integrating with Chat service endpoint

sacheu:
* do not add request.uri in the url template. 
* remove unnecessary properties in RequestViewModel that are not used
* Pull 'ContentType' header from request parameter. 
* Refactor a function to generate the required, optionals query params and headers
* Move logic to build optional header /query params to the unwrapped options code block.

### DIFF
--- a/src/AutorestSwift/View Models/KeyValueViewModel.swift
+++ b/src/AutorestSwift/View Models/KeyValueViewModel.swift
@@ -69,10 +69,6 @@ struct KeyValueViewModel {
                 // Convert into String in generated code
                 self.value = "String(\(param.name))"
             }
-        } else if key == "endpoint" {
-            self.value = "baseUrl.absoluteString" // baseUrl is a property of the generated client code
-            self.optional = false
-            self.paramName = nil
         } else {
             self.value = ""
             self.optional = false
--- a/src/AutorestSwift/View Models/KeyValueViewModel.swift
+++ b/src/AutorestSwift/View Models/KeyValueViewModel.swift
@@ -49,7 +49,7 @@ struct KeyValueViewModel {
         - Parameter operation: the operation which this paramter exists.
      */
     init(from param: Parameter, with operation: Operation) {
-        self.key = param.serializedName!
+        self.key = param.serializedName ?? param.name
 
         if let constantSchema = param.schema as? ConstantSchema {
             let isString: Bool = constantSchema.valueType.type == AllSchemaTypes.string
--- a/src/AutorestSwift/View Models/OperationViewModel.swift
+++ b/src/AutorestSwift/View Models/OperationViewModel.swift
@@ -82,12 +82,6 @@ struct OperationViewModel {
         self.name = operationName(for: operation.name)
         self.comment = ViewModelComment(from: operation.description)
 
-        var signatureComments: [String] = []
-        for param in operation.signatureParameters ?? [] where param.description != "" {
-            signatureComments.append("   - \(param.name) : \(param.description)")
-        }
-        self.signatureComment = ViewModelComment(from: signatureComments.joined(separator: "\n"))
-
         var params = OperationParameters()
         var pipelineContext = [KeyValueViewModel]()
 
@@ -153,6 +147,12 @@ struct OperationViewModel {
         signatureParams.forEach {
             signaturePropertyViewModel.append(ParameterViewModel(from: $0))
         }
+
+        var signatureComments: [String] = []
+        for param in signatureParams where param.description != "" {
+            signatureComments.append("   - \(param.name) : \(param.description)")
+        }
+        self.signatureComment = ViewModelComment(from: signatureComments.joined(separator: "\n"))
 
         self.signatureParams = signaturePropertyViewModel
 
--- a/src/AutorestSwift/View Models/OperationViewModel.swift
+++ b/src/AutorestSwift/View Models/OperationViewModel.swift
@@ -26,6 +26,41 @@
 
 import Foundation
 
+struct OperationParameters {
+    var header: Params
+    var query: Params
+    var path: [KeyValueViewModel]
+
+    init() {
+        self.header = Params()
+        self.query = Params()
+        self.path = [KeyValueViewModel]()
+    }
+
+    init(operationParameters: OperationParameters) {
+        self.header = operationParameters.header
+        self.query = operationParameters.query
+        self.path = operationParameters.path
+    }
+}
+
+struct Params {
+    // Query Params/Header in initializer
+    var required: [KeyValueViewModel]
+    // Query Params/Header need to add Nil check
+    var optional: [KeyValueViewModel]
+
+    init(from params: Params) {
+        self.required = params.required
+        self.optional = params.optional
+    }
+
+    init() {
+        self.required = [KeyValueViewModel]()
+        self.optional = [KeyValueViewModel]()
+    }
+}
+
 /// View Model for an operation.
 /// Example:
 ///     // a simple endpoint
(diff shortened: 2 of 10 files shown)


src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -69,10 +69,6 @@ struct KeyValueViewModel {
                 // Convert into String in generated code
                 self.value = "String(\(param.name))"
             }
-        } else if key == "endpoint" {
-            self.value = "baseUrl.absoluteString" // baseUrl is a property of the generated client code
-            self.optional = false
-            self.paramName = nil
```
tjprescott, line 75: Why did this go away?

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -63,27 +63,18 @@ struct OperationViewModel {
...
-        }
+        buildQueryParamsHeaders(
+            parameters: operation.parameters,
+            with: operation,
+            requiredQueryParams: &requiredQueryParams,
+            requiredHeaders: &requiredHeaders,
+            optionalQueryParams: &optionalQueryParams,
+            optionalHeaders: &optionalHeaders,
+            pathParams: &pathParams
+        )
```
tjprescott, line 77: Seems weird to have one parameter that is `with`. Might make more sense to just use `operation` like the others. The use of `inout` syntax is certainly atypical but is, I'm assuming, a result of using structs. A cleaner implementation might be to have a single utility struct that holds all the useful info, and then you can pass in the before and return the after:
```swift
struct OperationParameters {
   let header: Params
   let query: Params
   let path: Params
}

struct Params {
  let required: [String:String]
  let optional: [String:String]
}

let finalParams = buildQueryParamsHeaders(parameters:...,operation:..., params: params)
```

Then you could just dot into what you want `for param in params.header.required { ... }`

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -196,3 +192,32 @@ private func operationName(for operationName: String) -> String {
...
+
+/// Build a list of required and optional query params and headers from a list of parameters
+private func buildQueryParamsHeaders(
+    parameters: [Parameter]?,
+    with operation: Operation,
+    requiredQueryParams: inout [KeyValueViewModel],
+    requiredHeaders: inout [KeyValueViewModel],
+    optionalQueryParams: inout [KeyValueViewModel],
+    optionalHeaders: inout [KeyValueViewModel],
+    pathParams: inout [KeyValueViewModel]
```
tjprescott, line 204: Do these not have initial values?  If all we are doing is processing parameters into a structure that is more manageable, then I would recommend the approach I suggested above. Another alternative that is messier but avoids the use of `inout` (which is discouraged anyways) is to return a tuple.

templates/ModelFile.stencil
```diff
@@ -16,7 +16,7 @@ public {{ model.objectType }} {{ model.name }} : Codable {
     {% endfor %}
     public init(
     {% for property in model.properties %}
-        {{ property.name }}: {{ property.type }}{{ property.initDefaultValue }}{% ifnot forloop.last %},{% endif%}
+        {{ property.name }}: {{ property.type }} {{ property.initDefaultValue }}{% ifnot forloop.last %},{% endif%}
```
tjprescott, line 19: If we leave the trailing comma, will swiftlint or swiftformat clean it up, or does the compiler just fail? I would love to avoid having this conditional logic here is possible.

templates/OperationOptionsSnippet.stencil
```diff
@@ -1,14 +1,17 @@
 // Process endpoint options
 if let options = options {
     // Query options
-    {% for opt in op.options.queryOptions %}
-    if let {{ opt.name }} = options.{{ opt.name }} { queryParams.append("{{ opt.name }}", {{ opt.name}}) }
+    {% for param in op.optionalQueryParams %}
```
tjprescott, line 4: So with my above suggestion, this would look like `for param in op.params.query.optional`

templates/OperationOptionsSnippet.stencil
```diff
@@ -1,14 +1,17 @@
...
+    if let {{ param.paramName }} = options.{{ param.paramName }} {
+        queryParams.append("{{ param.key }}", {{ param.value}})
+    }
     {% endfor %}
 
     // Header options
-    {% for opt in op.options.headerOptions %}
-    if let {{ opt.name }} = options.{{ opt.name }} {
-        headers[.{{ opt.name }}] = {{ opt.name }}
+    {% for header in op.optionalHeaders %}
```
tjprescott, line 11: `op.params.header.optional`

templates/OperationUrlSnippet.stencil
```diff
@@ -1,7 +1,7 @@
 // Construct URL
-let urlTemplate = "{{ op.request.uri }}{{ op.request.path }}"
-let pathParams = [
-    {% for param in op.uriParams %}
+let urlTemplate = "{{ op.request.path }}"
+let pathParams:[String: String] = [
```
tjprescott, line 3: Is there a case where omitting this type causes problems? Even in the empty scenario, you provide two strings, so the compiler should be able to figure this out.

templates/OperationUrlSnippet.stencil
```diff
@@ -1,7 +1,7 @@
...
+    {% for param in op.pathParams %}
```
tjprescott, line 4: op.params.path

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: The service endpoint is from the 'baseUrl' which is initialized in the initializer of the Client code. The whole url will be resolved in ` guard let url = self.url(forTemplate: urlTemplate, withKwargs: pathParams) ` in the generated code. 

So I don't see we need to above binding between endpoint and baseUrl. 

In fact, the old implementation urltemplate ="{endpoint}{path}" will resolved in a wrong url. With the old implementation, the url was 'http://myserverhttp://myserver/{the path}'. so the {endpoint} is redundant that is why I remove it.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Will implement without `inout`

templates/ModelFile.stencil
sacheu: Agreed.  But swiftformat, swiftlint does not clean them up. Compile will fail if there is trailing comma in the parameter list of initializer.

templates/OperationUrlSnippet.stencil
sacheu: Removed.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -53,37 +82,14 @@ struct OperationViewModel {
...
-                viewModel.optional ? optionalHeaders.append(viewModel) : requiredHeaders
-                    .append(viewModel)
-            case .path,
-                 .uri:
-                uriParams.append(viewModel)
-            default:
-                continue
-            }
-        }
+        params = buildQueryParamsHeaders(
```
tjprescott, line 88: In this case, since params is initially empty, you can just use
```Swift
var params = buildQueryParamsHeaders(
  parameters: operation.parameters,
  operation: operations
)
```
Or, even better, just move the implementation of `buildQueryParamsHeaders` into an init for `OperationParameters()`.

```Swift
struct OperationParamters() {
  init(parameters: [Parameters], operation: Operation) {
    // move all that logic in here
  }
}
```

sacheu: There are 2 cases: 
1.  when all header and query are empty.
2. when header, query are populated and i want to add more header/query to existing entries in those arrays

If i move all the buildQueryParamsHeaders() logic to init(), I need to copy that to the 2nd init ` init(operationParameters: OperationParameters) {`, 

So that will result in duplicate code.

sacheu: buildQueryParamsHeaders is called in 2 places in the init() of OperationViewModel

tjprescott:
Fixes #89.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: I working on making this more concise.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Thanks for the suggestion. Changed per suggestion.

PR #114, opened 2020-08-05
Limited JSON Patch

### LINKED ISSUE #97: JSON Patch implementation
http://jsonpatch.com/

We will need to support JSON patch operations, which are not as simple as PUT operations. We will need to be able to distinguish between default `nil` values and the user explicitly setting a value to `nil`, as well as format the request body in accordance with the JSON PATCH schema.

tjprescott:
Closes #97.

This works as intended, but the generated code has bugs, so we should discuss what the approach should be: a) workaround in code-gen or b) fix in Swagger.

```Swift
// Value of optional type must be unwrapped to refer to member 'foo'...
if let val = body.foo {
  patch.replace(atPath: "foo/", withValue: val)
}
```
The problem here is that for a PATCH, the body should not be optional since there's no point in a patch operation without a body.

```Swift
// Initializer for conditional binding must have Optional type
if let val = body.bar {
  patch.replace(atPath: "bar/", withValue: val)
}
```
Here the body isn't optional, but neither is the property `bar`. All properties on the patch object should be optional, since it is the optionality that determines whether an update is sent. Arguably, since this object only contains one property, it should be required, but that's very special case. The question is, should we work around that in the code generator?


sacheu:
Since `All properties on the patch object should be optional`, so even there is only 1 property, it should still be optionals. 

So we can make this object contains 1 property to be optionals and that resolve this issue? Am I missing something?

src/AutorestSwift/Code Generator/SwiftGenerator.swift
```diff
@@ -52,10 +52,20 @@ class SwiftGenerator: CodeGenerator {
     func generate() throws {
         let modelUrl = baseUrl.with(subfolder: .models, withTargetName: targetName)
         let optionsUrl = baseUrl.with(subfolder: .options, withTargetName: targetName)
+        let utilUrl = baseUrl.with(subfolder: .util, withTargetName: targetName)
         try modelUrl.ensureExists()
         try optionsUrl.ensureExists()
+        try utilUrl.ensureExists()
         logger.log("Base URL: \(baseUrl.path)")
 
+        // Create PatchUtil.swift file
```
sacheu, line 61: Should we check if the code model has patch operation before generating this PatchUtil.swift file?

src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -26,35 +26,54 @@
...
         // load HttpRequest properties
         let httpRequest = request.protocol.http as? HttpRequest
         self.path = httpRequest?.path ?? ""
         self.method = httpRequest?.method.rawValue ?? ""
 
-        // load HttpWithBodyRequest specfic properties
-        let httpWithBodyRequest = request.protocol.http as? HttpWithBodyRequest
-
-        // TODO: only support the first signature parameter in Reqest now
+        // TODO: only support the first signature parameter in Request now
```
sacheu, line 65: This comment can be removed too.

src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -26,35 +26,54 @@
...
         let bodyParams = getBodyParameters(signatureParameters: request.signatureParameters)
         assert(bodyParams.count <= 1, "Unexpectedly found more than 1 body parameters in request... \(request.name)")
-
         let bodyParam = bodyParams.first
-        self.objectType = bodyParam?.schema.name
-        if belongsInSignature(param: bodyParam) {
```
sacheu, line 52: So if we don't check whether the 'body' is from signature or from options, the function  belongsInSignature() can be removed

templates/PatchUtilFile.stencil
```diff
@@ -0,0 +1,70 @@
+import Foundation
+
+// TODO: Remove this when these objects are in AzureCore
```
sacheu, line 3: Do we need PatchUtilFile.swift ? I think we only need PatchUtilFile.stencil for generating PatchUtilFile.swift in generated code.

templates/OperationRequestPatchBodySnippet.stencil
```diff
@@ -0,0 +1,21 @@
+// Construct patch request
+let patch = MergePatchObject()
+{% for name in op.request.bodyParam.propertyNames %}
+if let val = {{ op.request.bodyParam.name }}.{{ name }} {
+    patch.replace(atPath: "{{name}}/", withValue: val)
```
sacheu, line 5: so we only support replace now? others like 'add', 'remove' is not going to be supported?

sacheu: Where in swagger/code model can tell us whether it is a 'patch.replace' instead of 'patch.add' or 'patch.remove'?

src/AutorestSwift/Code Generator/SwiftGenerator.swift
tjprescott: Shouldn't need to since it is going in the Util folder and not the Models folder. What we might need want to check is whether there is something in Models that shares the same name as something in PatchUtil... but since this is temporary anyways, we are probably okay.

src/AutorestSwift/View Models/RequestViewModel.swift
tjprescott: I'll replace it with an assert.

src/AutorestSwift/View Models/RequestViewModel.swift
tjprescott: I thought the logic for options was anything that was a query or header param, which means we wouldn't expect to ever find the body param in the options?

templates/PatchUtilFile.stencil
tjprescott: Hmmm... good point. I think I had it in there so I could test it :)

src/AutorestSwift/View Models/RequestViewModel.swift
sacheu: Yes. you are right.  belongsInSignature can be removed.

templates/OperationRequestPatchBodySnippet.stencil
tjprescott: All of the methods are in the PatchUtil file, but I don't see any of the actual SDK patch scenarios using anything other than `replace`. To use any of the others, we'd need to know what the original object looked like and diff them, which I would really not want to do.

src/AutorestSwift/View Models/RequestViewModel.swift
tjprescott: And there already is one 😅

templates/PatchUtilFile.stencil
tjprescott: originally it was in there because I envisioned creating the patch object in the generator and only outputting the result to the view model. Then we wouldn't need to copy this file to the generated code. But copying would result in fewer changes when AzureCore supports PATCH, so I went that route. I've removed this file.

tjprescott: And by "this file" I mean PatchUtil.swift in the autorest.swift repo :)

PR #122, opened 2020-08-06
Update method names for API calls with body params

### LINKED ISSUE #112: Change how body parameters are represented in method signatures
Our code gen currently produces signature like this:

```Swift
updateCat(id: String, body: UpdateCatRequest, withOptions: UpdateCatOptions? = nil, completionHandler: ...) {
```

Instead, it should look like this:
```Swift
update(cat: UpdateCatRquest, id: String, withOptions: UpdateCatOptions? = nil, completionHandler: ...) {
```

This will involve:
- always listing the body parameter (if any) first in the signature
- inferring the name of the body type from the method or Options object name
- stripping the name of the body type from the method name

tjprescott:
Closes #112.

### DIFF
--- a/src/AutorestSwift/View Models/OperationViewModel.swift
+++ b/src/AutorestSwift/View Models/OperationViewModel.swift
@@ -82,6 +82,7 @@ struct OperationViewModel {
     let name: String
     let comment: ViewModelComment
     let signatureComment: ViewModelComment
+    let bodyParam: ParameterViewModel?
     let signatureParams: [ParameterViewModel]
     let returnType: ReturnTypeViewModel?
     let params: OperationParameters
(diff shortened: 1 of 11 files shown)


src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -102,30 +103,33 @@ struct OperationViewModel {
...
+        )
+        guard let request = operation.requests?.first else { fatalError("No requests found.") }
+        requests.append(RequestViewModel(from: request, with: operation))
 
-            let requestSignatureParams = filterParams(for: request.signatureParameters, with: [.path, .uri, .body])
-            let requestOptionsParams = filterParams(
-                for: request.signatureParameters,
-                with: [.header, .query]
-            )
+        let requestSignatureParams = filterParams(for: request.signatureParameters, with: [.path, .uri])
```
sacheu, line 119: so you remove . body from the requestSignaureParams since it is handled specially (put that as first of the signature)

tjprescott: Exactly. I think swift arrays are ordered, but this guarantees we can treat any body params first.

PR #124, opened 2020-08-06
Add custom init for decoder and encoder

### LINKED ISSUE #118: Need custom decoder (and probably encoder?) to deserialize Date
To property deserialize Date objects in our models, we can't rely on the default Codable implementation for Date, as it only works if the dates are sent as Double. For our situation, where the Date arrives as a string, we need custom decoding logic, as shown below:

```Swift
    public init(from decoder: Decoder) throws {
        let container = try decoder.container(keyedBy: CodingKeys.self)
        // most properties will be straightforward like this...
        self.id = try? container.decode(String.self, forKey: .id)

        // date objects will need to be generated like this
        if let dateString = try? container.decode(String.self, forKey: .someDateValue) {
            let dateFormatter = DateFormatter()
            dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSSZ"
            self.someDateValue = dateFormatter.date(from: dateString)
        } else {
            self.someDateValue = nil
        }
    }
```

It's entirely possible we'll need similar logic for encoding.

sacheu:
Closes #118.

### DIFF
--- a/templates/ModelDecoderEncoderSnippet.stencil
+++ b/templates/ModelDecoderEncoderSnippet.stencil
@@ -10,9 +10,9 @@ public init(from decoder: Decoder) throws {
     let container = try decoder.container(keyedBy: CodingKeys.self)
     {% for property in model.properties %}
         {% if property.isDate %}
-        {% include ModelDecoderDateProperty.stencil %}
+        {% include "ModelDecoderDateProperty.stencil" %}
         {% else %}
-        {% include ModelDecoderProperty.stencil %}
+        {% include "ModelDecoderProperty.stencil" %}
         {% endif %}
     {% endfor %}
 }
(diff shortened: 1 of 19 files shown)


templates/ModelDecoderEncoderSnippet.stencil
```diff
@@ -0,0 +1,31 @@
+{% if model.hasDateProperty %}
```
tjprescott, line 1: Truthfully it is probably okay to just always generate this. I suspect there will be other things than just dates that will require some custom encode/decode logic.

templates/ModelDecoderEncoderSnippet.stencil
```diff
@@ -0,0 +1,31 @@
...
+public func encode(to encoder: Encoder) throws {
+    var container = encoder.container(keyedBy: CodingKeys.self)
+    {% for property in model.properties %}
+    {% if property.isDate %}
+    {% include "ModelEncoderDateProperty.stencil" %}
+    {% else %}
+    {% include "ModelEncoderProperty.stencil" %}
+    {% endif %}
+    {% endfor %}
+}
```
tjprescott, line 29: It would probably be okay to split this out into a single ModelDecoderProperty file that works like a big switch. I doubt any individual type would need more than 10 LOC.

tjprescott:
Now includes the changes from #125.

PR #128, opened 2020-08-10
Add Jazzy template for generated code

### LINKED ISSUE #123: Generate a jazzy.yaml file
We need to generate a jazzy.yaml config file to permit generation of docs from the Swift docstrings. See: 
https://azure.github.io/azure-sdk/ios_documentation.html#swift-docstrings

ping @bsiegel as he did most of the docs generation for `azure-sdk-for-ios`

tjprescott:
Closes #123.

This only works for packages built into the `azure-sdk-for-ios` repo.


templates/JazzyFile.stencil
```diff
@@ -0,0 +1,22 @@
+author: Microsoft
+author_url: https://azure.github.io/azure-sdk/
+github_url: https://github.com/Azure/azure-sdk-for-ios
+module: {{ model.name }}
+module_version: 0.1
```
sacheu, line 5: Do we need to pull this value from swagger file 's API version?

templates/JazzyFile.stencil
```diff
@@ -0,0 +1,22 @@
...
+sdk: iphonesimulator
+xcodebuild_arguments:
+  - "-workspace"
+  - "AzureSDK.xcworkspace"
+  - "-scheme"
+  - "{{ model.name }}"
+  - "-sdk"
+  - "iphonesimulator"
+exclude:
+  - "*/Source/ObjectiveCCompatibility/*"
```
sacheu, line 21: we probably don't need these 2 excludes , right?

templates/JazzyFile.stencil
tjprescott: it won't be the swagger's API versions, but it does need to be dynamic.

templates/JazzyFile.stencil
```diff
@@ -0,0 +1,22 @@
...
+readme: ../sdk/{{ model.namespace }}/{{ model.name }}/README.md
+skip_undocumented: false
+hide_unlisted_documentation: false
+theme: fullwidth
+output: ../build/jazzy/{{ model.name }}
+clean: true
+sdk: iphonesimulator
+xcodebuild_arguments:
+  - "-workspace"
+  - "AzureSDK.xcworkspace"
```
sacheu, line 15: The generated code does not have a workspace generated, Will `"AzureSDK.xcworkspace"` work?

templates/JazzyFile.stencil
tjprescott: I kept them in because they were taken from storage.

templates/JazzyFile.stencil
tjprescott: I'm trying to test that now. Right now the model is that we generate into the azure-sdk-for-ios repo, so it should eventually land in that workspace.

sacheu:
So now that jazzy config file is generated, do you need to run jazzy as p so that the doc files will get created as well?

PR #130, opened 2020-08-11
Disable swiftlint rule in generated code

sacheu:
With this PR, there should be no lint error in the generated code. Only in the dependency code.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,2 +1,5 @@
 disabled_rules: # rule identifiers to exclude from running
 - large_tuple
+- function_body_length
+
+excluded: # paths to ignore during linting.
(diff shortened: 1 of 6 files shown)


.swiftlint.yml
```diff
@@ -1,2 +1,5 @@
 disabled_rules: # rule identifiers to exclude from running
 - large_tuple
+- function_body_length
```
tjprescott, line 3: So why `function_body_length` here instead of in each file? This would apply even if someone decided to hand write some extension code, so putting it in the individual files is probably best.

sacheu: This is to disable the lint warning for the AutoRest swift project. Not the generated code.

sacheu: And we do have // swiftlint:disable function_body_length in the ServiceClientFile stencil.

tjprescott: See I was mistaken :D

PR #134, opened 2020-08-11
Use Stencil smart trim behavior to remove excess empty lines in gener…

sacheu:
…ated code.

Since the PR is not merged to master, we need to change Stencil dependency to the PR branch instead.

Per PR comment: 
smart Trim behavior means `It also comes with a built in .smart case that removes whitespace before a block, and whitespace and a single newline after a block.`


sacheu:
https://github.com/stencilproject/Stencil/pull/287 is the Stencil PR

PR #143, opened 2020-08-18
RPC Changes

tjprescott:
This doesn't compile because `call(method:params:)` isn't implemented.

### DIFF
--- a/src/AutorestSwift/Code Generator/AutoRestHandler.swift
+++ b/src/AutorestSwift/Code Generator/AutoRestHandler.swift
@@ -3,35 +3,51 @@ import NIO
 import os.log
 
 class AutoRestHandler {
-    private let logger: Logger
+    // MARK: Properties
 
-    init() {
-        self.logger = Logger(withFileName: "autorest-swift-debug.log")
+    private let plugin: AutoRestPlugin
+    private var logger: Logger {
+        return plugin.logger
     }
 
-    func handle(method: String, params _: RPCObject, callback: (RPCResult) -> Void) {
+    // MARK: Initializers
+
+    public init(withPlugin plugin: AutoRestPlugin) {
+        self.plugin = plugin
+    }
+
+    // MARK: Methods
+
+    func handle(method: String, params: RPCObject, callback: (RPCResult) -> Void) {
         switch method.lowercased() {
         case "getpluginnames":
-            logger.logToURL("Get a getpluginnames request")
-            return callback(.success(.list([.string("swift")]), RPCObjectType.response, ""))
-
+            callback(
+                .success(.list([.string("swift")]), .response, "")
+            )
         case "process":
-            logger.logToURL("Get a process request")
-
-            let sessionId = "session_7"
-
-            logger.logToURL("Send back a process response")
-            callback(.success(.bool(true), RPCObjectType.response, ""))
-
-            logger.logToURL("Send a ListInputs request")
-            return callback(.success(
-                .list([.string(sessionId), .string("code-model-v4")]),
-                RPCObjectType.request,
-                "ListInputs"
-            ))
+            plugin.sessionId = params.asList?.last?.asString ?? ""
+            callback(
+                .success(.bool(true), .response, "")
+            )
+            initializationComplete()
         default:
             callback(.failure(RPCError(.invalidMethod)))
             logger.logToURL("invalid method: \(method)")
         }
+    }
+
+    func initializationComplete() {
+        guard let sessionId = plugin.sessionId else { fatalError() }
+        let listInputsRequest: RPCObject = .list([.string(sessionId), .string("code-model-v4")])
+        plugin.call(method: "ListInputs", params: listInputsRequest)
+        // TODO: Retrieve the code model filename
+        let filename = "code-model-v4.yaml"
+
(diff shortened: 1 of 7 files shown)


src/AutorestSwift/Code Generator/AutoRestHandler.swift
```diff
@@ -3,35 +3,51 @@ import NIO
...
 
 class AutoRestHandler {
-    private let logger: Logger
+    // MARK: Properties
 
-    init() {
-        self.logger = Logger(withFileName: "autorest-swift-debug.log")
+    private let plugin: AutoRestPlugin
+    private var logger: Logger {
+        return plugin.logger
```
tjprescott, line 10: This was to eliminate duplication, since the handler can reach back to its parent plugin.

src/AutorestSwift/Code Generator/AutoRestHandler.swift
```diff
@@ -3,35 +3,51 @@ import NIO
...
+        case "getpluginnames":
+            callback(
+                .success(.list([.string("swift")]), .response, "")
+            )
+        case "process":
+            plugin.sessionId = params.asList?.last?.asString ?? ""
+            callback(
+                .success(.bool(true), .response, "")
+            )
+            initializationComplete()
```
tjprescott, line 32: I don't like calling this here. Would rather have some kind of semaphore somewhere that we can trip here to signal that all of the messages we expect to come from Autorest have come.... but since there's only one process message, this should be okay?

src/AutorestSwift/Code Generator/AutoRestHandler.swift
```diff
@@ -3,35 +3,51 @@ import NIO
...
-                "ListInputs"
-            ))
+    func handle(method: String, params: RPCObject, callback: (RPCResult) -> Void) {
+        switch method.lowercased() {
+        case "getpluginnames":
+            callback(
+                .success(.list([.string("swift")]), .response, "")
+            )
+        case "process":
+            plugin.sessionId = params.asList?.last?.asString ?? ""
```
tjprescott, line 28: I added the various `asFoo` calculated parameters to `RPCObject` to facilitate this.

src/AutorestSwift/Code Generator/AutoRestHandler.swift
```diff
@@ -3,35 +3,51 @@ import NIO
...
         default:
             callback(.failure(RPCError(.invalidMethod)))
             logger.logToURL("invalid method: \(method)")
         }
     }
+
+    func initializationComplete() {
+        guard let sessionId = plugin.sessionId else { fatalError() }
+        let listInputsRequest: RPCObject = .list([.string(sessionId), .string("code-model-v4")])
+        plugin.call(method: "ListInputs", params: listInputsRequest)
```
tjprescott, line 42: I don't think we can just use the handler to send requests. It seems we need a method to initiate a request.

src/AutorestSwift/Code Generator/AutoRestPlugin.swift
```diff
@@ -3,25 +3,44 @@ import NIO
...
-        logger.logToURL("Start NIO pipe")
+    var logger: Logger!
+    var sessionId: String?
 
-        let channel = NIOPipeBootstrap(group: MultiThreadedEventLoopGroup(numberOfThreads: 5))
+    public func start() throws -> EventLoopFuture<AutoRestPlugin> {
+        logger = Logger(withFileName: "autorest-swift-debug.log")
+        logger.logToURL("Starting NIO pipeline")
+        let handler = AutoRestHandler(withPlugin: self)
+        let channel = NIOPipeBootstrap(group: MultiThreadedEventLoopGroup(numberOfThreads: 1))
```
tjprescott, line 13: Made this one thread so logging statements don't get interleaved. This should make comprehending the logs easier.

src/AutorestSwift/Code Generator/jsonrpc/Model.swift
```diff
@@ -326,6 +326,60 @@ public enum RPCObject: Equatable {
...
+    }
+
+    public var asDict: [String: RPCObject]? {
+        switch self {
+        case let .dictionary(value):
+            return value
+        default:
+            return nil
+        }
+    }
```
tjprescott, line 382: These needed to work with `RPCObject` in anything resembling a reasonable fashion.

src/AutorestSwift/Code Generator/AutoRestPlugin.swift
```diff
@@ -3,25 +3,44 @@ import NIO
...
+//         }
+//         let promise: EventLoopPromise<JSONResponse> = channel.eventLoop.makePromise()
+//         let request = JSONRequest(id: NSUUID().uuidString, method: method, params: JSONObject(params))
+//         let requestWrapper = JSONRequestWrapper(request: request, promise: promise)
+//         let future = channel.writeAndFlush(requestWrapper)
+//         future.cascadeFailure(to: promise) // if write fails
+//         return future.flatMap {
+//             promise.futureResult.map { Result($0) }
+//         }
+    }
```
tjprescott, line 45: I think we need a call method on the plugin to send events, as described in:
https://github.com/apple/swift-nio-examples/blob/master/json-rpc/Sources/JsonRpc/Client.swift

PR #145, opened 2020-08-20
RPC Refactor

tjprescott:


src/AutorestSwift/JsonRpc/Client.swift
```diff
@@ -0,0 +1,225 @@
...
+                    CodableCodec<JSONResponse, JSONRequest>(),
+                    name: "AutorestClient"
+                ).flatMap {
+                    context.pipeline.addHandler(
+                        Handler(self.initalizationComplete),
+                        name: "ClientHandler"
+                    )
+                }
+            }
+        }
```
tjprescott, line 71: Why do we need to remove the server handler? If it has both the client and server handlers simultaneously, what happens?

src/AutorestSwift/JsonRpc/Client.swift
```diff
@@ -0,0 +1,225 @@
...
+
+// TODO: Update license. Reproduced from: https://github.com/apple/swift-nio-examples/tree/master/json-rpc/Sources/JsonRpc
+
+import Foundation
+import NIO
+
+public final class ChannelClient {
+    private let group: MultiThreadedEventLoopGroup
+    public let config: Config
+    private var context: ChannelHandlerContext?
```
tjprescott, line 35: Is there a way to not require this? If the ChannelClient/Server have a link back to the AutorestPlugin, can they access the context that way?

src/AutorestSwift/JsonRpc/Server.swift
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+private class Handler: ChannelInboundHandler, RemovableChannelHandler {
+    public typealias InboundIn = JSONRequest
+    public typealias OutboundOut = JSONResponse
+
+    private let closure: RPCClosure
+
+    private let logger = FileLogger(withFileName: "autorest-swift-debug.log")
```
tjprescott, line 107: Perhaps we could have a static logger on AutorestPlugin or just a singleton Logger so we don't need to have multiple logger properties?

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
```diff
@@ -98,7 +98,16 @@ class AutorestPlugin {
...
+            guard let sessionId = params.asList?.last?.asString else {
+                logger.log("Process request get invalid session id")
+                callback(.failure(RPCError(
+                    kind: .invalidParams,
+                    description: "Process request get invalid session id"
+                )))
+                return
+            }
+
+            plugin.sessionId = sessionId
```
tjprescott, line 110: Since this is not a "production level" tool, we could also just declare `var sessionId: String!`.  Then you could keep the original, more concise usage.

```Swift
plugin.sessionId = params.asList?.last?.asString

print(plugin.sessionId) // automatically unwrapped
```

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
```diff
@@ -129,7 +138,11 @@ class AutorestPlugin {
     }
 
     func handleListInputs(response: RPCObject) {
-        let filename = response.asList?.first?.asString ?? ""
+        guard let filename = response.asList?.first?.asString else {
+            logger.log("handleListInputs filename is nil")
+            return
```
tjprescott, line 143: We can probably just make this fatalError with the same message. If this fails, logically the whole process should fail.

src/AutorestSwift/JsonRpc/Client.swift
sacheu: If we leave both handlers simultaneously, the server handler will crash when we are  operating as a client. And vice versa. For exampole, in Server mode, we have this handler`CodableCodec<JSONRequest, JSONResponse >()`, it expects input as 'JSONResponse' which is not what it expects 'JSONRequest', it will cause casting error in handling the message.

src/AutorestSwift/JsonRpc/Client.swift
sacheu: The issue is we need the context of the Handler in Client.swift, not any context of any handler. 
That is why this property is set after 'Handler in Client.swift' is added (see handlerAdded of Handler in Client.swift'

src/AutorestSwift/JsonRpc/Client.swift
```diff
@@ -30,70 +30,82 @@ import Foundation
...
+    private let group: MultiThreadedEventLoopGroup
     public let config: Config
-    private var channel: Channel?
-
-    public init(group: MultiThreadedEventLoopGroup, config: Config = Config()) {
+    private var context: ChannelHandlerContext?
+    private let processCallback: ProcessCallback
+    private let logger = FileLogger(withFileName: "autorest-swift-debug.log")
+    // TODO: get this value based on last response id from AutoRest
+    var id = 3
```
tjprescott, line 39: What is this used for?

src/AutorestSwift/JsonRpc/Server.swift
sacheu: Changed in new iteration

src/AutorestSwift/JsonRpc/Client.swift
sacheu: This is the initial value for the request id when we send message (as a client ) to Autorest.  We should fix it to retrieve it programmically (this value should be the id of the last message we get from Autorest when we were acting as Server.

We need to increment this value (id of the request) for each request when we send message to Autorest.

src/AutorestSwift/Code Generator/Logger.swift
```diff
@@ -67,6 +67,8 @@ class FileLogger {
 
     let url: URL
 
+    static let instance = FileLogger(withFileName: "autorest-swift-debug.log")
```
tjprescott, line 70: Typically in iOS lingo this would be called `shared`.

src/AutorestSwift/JsonRpc/Client.swift
sacheu: Wil open a new PR to address this. Don't want this PR to be so big.

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
```diff
@@ -98,22 +93,13 @@ class AutorestPlugin {
...
-
-            plugin.sessionId = sessionId
+            plugin.sessionId = params.asList?.last?.asString
             startChannelClient(context: context)
         // TODO: Sending back Process response will stop Autorest.
         // callback(.success(.bool(true)))
         default:
             callback(.failure(RPCError(kind: .invalidMethod, description: "Incoming Handler get invalid method")))
-            logger.log("invalid method: \(method)")
+            FileLogger.instance.log("invalid method: \(method)")
```
tjprescott, line 102: Similarly here. If the method is invalid, we should fatalError the process. This is basically true of any failure in the case of this particular tool.

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
```diff
@@ -139,8 +125,8 @@ class AutorestPlugin {
 
     func handleListInputs(response: RPCObject) {
         guard let filename = response.asList?.first?.asString else {
-            logger.log("handleListInputs filename is nil")
-            return
+            FileLogger.instance.log("handleListInputs filename is nil")
+            fatalError("handleListInputs filename is nil")
```
tjprescott, line 129:
```suggestion
            msg = "Unable to retrieve filename from Autorest response."
            FileLogger.instance.log(msg)
            fatalError(msg)
```

Alternatively, maybe we have a `logAndFail(msg)` helper function that always echoes the failure to the logger and then kills the process.

src/AutorestSwift/JsonRpc/Client.swift
```diff
@@ -71,12 +71,12 @@ public final class ChannelClient {
         }
 
         future.whenFailure { error in
-            self.logger.log("Switch to client mode failed: \(error)")
+            FileLogger.instance.log("Switch to client mode failed: \(error)")
```
tjprescott, line 74: Same. This should fail.

src/AutorestSwift/Code Generator/Logger.swift
sacheu: Renamed.

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
sacheu: Good idea.

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
sacheu: Fixed

PR #147, opened 2020-08-20
Refactor generator phase to work with RPC logic

tjprescott:
Modifies Manager to work with the RPC format of Autorest.

### DIFF
--- a/src/AutorestSwift/Code Generator/Logger.swift
+++ b/src/AutorestSwift/Code Generator/Logger.swift
@@ -71,11 +71,14 @@ class FileLogger {
 
     // MARK: Initializers
 
-    init(withFileName name: String) {
+    internal init(withFileName name: String, deleteIfExists: Bool = true) {
         guard let documentsUrl = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first else {
             fatalError("Unable to locate Documents directory.")
         }
         self.url = documentsUrl.appendingPathComponent(name)
+        if deleteIfExists {
+            try? FileManager.default.removeItem(atPath: url.path)
+        }
     }
 
     // MARK: Methods
@@ -87,8 +90,9 @@ class FileLogger {
         try? url.append(line: msg)
     }
 
-    func logAndFail(_ message: @autoclosure @escaping () -> String?) {
-        log(message())
-        fatalError(message() ?? "")
+    func logAndFail(_ message: @autoclosure @escaping () -> String?) -> Never {
+        let msg = message() ?? ""
+        log(msg)
+        fatalError(msg)
     }
 }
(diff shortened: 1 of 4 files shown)


src/AutorestSwift/Code Generator/Logger.swift
```diff
@@ -87,8 +90,9 @@ class FileLogger {
         try? url.append(line: msg)
     }
 
-    func logAndFail(_ message: @autoclosure @escaping () -> String?) {
-        log(message())
-        fatalError(message() ?? "")
+    func logAndFail(_ message: @autoclosure @escaping () -> String?) -> Never {
```
tjprescott, line 93: The `Never` return type avoids you having to `return` in guard statements when this is used.
```Swift
// THIS
guard let foo = bar else {
  logAndFail(...)
}

// Instead of THIS
guard let foo = bar else {
  logAndFail(...)
  return
}
```

src/AutorestSwift/Code Generator/Manager.swift
```diff
@@ -54,17 +66,32 @@ class Manager {
         }
     }
 
+    /// Initialize with a raw YAML string (i.e. the way it is received from Autorest).
+    init(withString string: String) {
+        self.mode = .autorest
+        self.inputString = Manager.sanitize(yaml: string)
+        self.destinationRootUrl = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
```
tjprescott, line 73: Creates a temporary destination URL for generating code. This way we can run our linters and such on this and then copy it all using Autorest WriteFile commands.

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
```diff
@@ -140,12 +138,19 @@ class AutorestPlugin {
...
-        let codeModel = response.asString ?? ""
-        // TODO: Now convert to code model and generate
+        guard let codeModel = response.asString else {
+            FileLogger.shared.logAndFail("Unable to retrieve code model from Autorest.")
+        }
+        let manager = Manager(withString: codeModel)
+        do {
+            try manager.run()
+        } catch {
+            FileLogger.shared.logAndFail("Code generation failure: \(error)")
```
tjprescott, line 153: Currently this fails for me because of Stencil files.

src/AutorestSwift/main.swift
```diff
@@ -25,19 +25,21 @@
 // --------------------------------------------------------------------------
 
 import Foundation
+import NIO
```
sacheu, line 28: Do we need this import here?

tjprescott: It would seem not!

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
sacheu: It works on my machine.

sacheu:
PR looks good. it works on my machine.

PR #155, opened 2020-08-24
Remove unnecessary debug logs

sacheu:


src/AutorestSwift/JsonRpc/Client.swift
```diff
@@ -184,15 +181,12 @@ private class Handler: ChannelInboundHandler, ChannelOutboundHandler, RemovableC
...
-        SharedLogger.debug("Client Handler handlerAdded")
         initComplete(context)
     }
 
     public func errorCaught(context: ChannelHandlerContext, error: Error) {
-        SharedLogger.debug("Client Handler errorCaught")
-        if let remoteAddress = context.remoteAddress {
-            print("server", remoteAddress, "error", error)
-        }
+        SharedLogger.error("Client Handler errorCaught")
```
tjprescott, line 188: should we log what the error is?

src/AutorestSwift/JsonRpc/Server.swift
```diff
@@ -124,7 +123,7 @@ private class Handler: ChannelInboundHandler, RemovableChannelHandler {
     }
 
     public func errorCaught(context: ChannelHandlerContext, error: Error) {
-        SharedLogger.debug("Server Handler errorCaught")
+        SharedLogger.error("Server Handler errorCaught")
```
tjprescott, line 126: Same comment. Should we log the error contents?

sacheu: Good point.

PR #158, opened 2020-08-24
Add preview release pipeline

daviwil:
This change adds a new pipeline configuration for publishing preview releases of @autorest/swift to GitHub Releases.

This pipeline is meant to be executed manually via the pipeline page: https://dev.azure.com/azure-sdk/internal/_build?definitionId=1938&_a=summary

### DIFF
--- a/eng/pipelines/publish-dev-release.yml
+++ b/eng/pipelines/publish-dev-release.yml
@@ -1,0 +1,22 @@
+trigger: none
+pr: none
+
+jobs:
+  - job: PreviewRelease
+
+    pool:
+      vmImage: 'macOS-10.15'
+
+    steps:
+      - script: sudo xcode-select --switch /Applications/Xcode_11.5.app
+        displayName: 'Use Xcode 11.5'
+
+      - script: npm install
+        displayName: Build
+
+      - script : |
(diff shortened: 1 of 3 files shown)


eng/pipelines/publish-dev-release.yml
```diff
@@ -0,0 +1,22 @@
...
+
+    steps:
+      - script: sudo xcode-select --switch /Applications/Xcode_11.5.app
+        displayName: 'Use Xcode 11.5'
+
+      - script: npm install
+        displayName: Build
+
+      - script : |
+          export DEV_VERSION=$(node -p -e "require('./package.json').version")-dev.$BUILD_BUILDNUMBER
```
sacheu, line 18: How often we will get a new build (a tgz file) with this dev pipeline?   Everytime there is a checkin to autorest swfit repo?
And how will the release (non dev version) of the pipeline work? how often a new build will be published in the release pipeline?

daviwil: You will be running this manually when you want to cut a preview release, same for the `npm` pipeline once we add it.  This is the process the other language teams use since shipping preview releases on every commit is pretty noisy.

PR #160, opened 2020-08-25
Body param flattening

### LINKED ISSUE #153: Remove casing logic
Casing logic is built into modeler four that is output by autorest, so we can remove the custom casing logic built into the plugin if the result is the same.

### LINKED ISSUE #159: Support `x-ms-client-flattening` for Body Parameters.
Will allow service teams to control the API signature through use of the `x-ms-client-flattening` extension in Swagger. 

Support `model-flattening.json` tests.

**Passing Tests**
- [x] test_resourceFlattening_getArray (#221)
- [x] test_resourceFlattening_putArray (#221)
- [x] test_resourceFlattening_getDictionary (#221)
- [ ] test_resourceFlattening_putDictionary
- [x] test_resourceFlattening_getResourceCollection (#221)
- [ ] test_resourceFlattening_putResourceCollection
- [ ] test_resourceFlattening_getWrappedArray
- [ ] test_resourceFlattening_putWrappedArray
- [ ] test_resourceFlattening_putSimpleProduct
- [ ] test_resourceFlattening_putSimpleProductWithGroupingFlattenParameterGroup
- [ ] test_resourceFlattening_postFlattenedSimpleProduct

tjprescott:
Closes #153. Closes #159.

Work to be done:

- [x] Resolve *where* optional body params should be located
- [ ] Update stencil to generate appropriate code for constructing the body param from virtual params

### DIFF
--- a/src/AutorestSwift/JsonRpc/AutorestPlugin.swift
+++ b/src/AutorestSwift/JsonRpc/AutorestPlugin.swift
@@ -166,6 +166,7 @@ class AutorestPlugin {
             SharedLogger.fail("Unable to retrieve code model from Autorest.")
         }
         let manager = Manager(withString: codeModel)
+        manager.saveCodeModel()
         do {
             try manager.run()
 
--- a/src/AutorestSwift/JsonRpc/AutorestPlugin.swift
+++ b/src/AutorestSwift/JsonRpc/AutorestPlugin.swift
@@ -166,7 +166,6 @@ class AutorestPlugin {
             SharedLogger.fail("Unable to retrieve code model from Autorest.")
         }
         let manager = Manager(withString: codeModel)
-        manager.saveCodeModel()
         do {
             try manager.run()
 
--- a/src/AutorestSwift/Models/Complex/Parameter.swift
+++ b/src/AutorestSwift/Models/Complex/Parameter.swift
@@ -27,7 +27,7 @@
 import Foundation
 
 /// A definition of an discrete input for an operation
-class Parameter: Value {
+class Parameter: Value, CustomDebugStringConvertible {
     /// suggested implementation location for this parameter
     let implementation: ImplementationLocation?
 
@@ -69,5 +69,15 @@ class Parameter: Value {
         if flattened != nil { try container.encode(flattened, forKey: .flattened) }
 
         try super.encode(to: encoder)
+    }
+
+    public static var codingKeys: CodingKeys.Type {
+        return CodingKeys.self
+    }
+
+    // MARK: CustomDebugStringConvertible
+
+    public var debugDescription: String {
+        return debugString() ?? description
     }
 }
--- a/src/AutorestSwift/Models/Complex/Parameter.swift
+++ b/src/AutorestSwift/Models/Complex/Parameter.swift
@@ -71,10 +71,6 @@ class Parameter: Value, CustomDebugStringConvertible {
         try super.encode(to: encoder)
     }
 
-    public static var codingKeys: CodingKeys.Type {
-        return CodingKeys.self
-    }
-
     // MARK: CustomDebugStringConvertible
 
     public var debugDescription: String {
--- a/src/AutorestSwift/Models/Complex/ParameterType.swift
+++ b/src/AutorestSwift/Models/Complex/ParameterType.swift
@@ -1,0 +1,115 @@
+// --------------------------------------------------------------------------
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+//
+// The MIT License (MIT)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the ""Software""), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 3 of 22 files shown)


src/AutorestSwift/Utils/String+Extensions.swift
```diff
@@ -77,23 +77,23 @@ extension String {
...
-    var toCamelCase: String {
-        let components = splitAndJoinAcronyms
-        var casedComponents = [String]()
-        for (index, comp) in components.enumerated() {
-            casedComponents.append(index == 0 ? comp : comp.capitalized)
-        }
-        let value = casedComponents.joined()
-        return value
-    }
+//    var toCamelCase: String {
```
sacheu, line 80: Maybe just deleted the code instead of commented out?

src/AutorestSwift/Utils/String+Extensions.swift
```diff
@@ -77,23 +77,23 @@ extension String {
...
+//    }
 
     /// Converts a string into one with camel casing. Acronyms are forced to camel case as well.
-    var toPascalCase: String {
-        let components = splitAndJoinAcronyms
-        let casedComponents = components.map { $0.capitalized }
-        let value = casedComponents.joined()
-        return value
-    }
+//    var toPascalCase: String {
```
sacheu, line 91: Maybe just deleted the code instead of commented out?

src/AutorestSwift/Code Generator/Manager.swift
```diff
@@ -216,4 +216,13 @@ class Manager {
             SharedLogger.fail("Fail to run tool \(tool).")
         }
     }
+
+    func saveCodeModel() {
+        guard let documentsUrl = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first else {
```
sacheu, line 221: why we need to save the code model to yaml file? is that only for debugging? How will this file get removed?

src/AutorestSwift/Models/Complex/ParameterType.swift
```diff
@@ -0,0 +1,115 @@
...
+        } else {
+            fatalError("Unable to decode ParameterType")
+        }
+    }
+
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.unkeyedContainer()
+        switch self {
+        case let .regular(param):
+            try container.encode(param)
```
sacheu, line 49: since both cases are doing the same thing, do we need a switch statement here?

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -213,15 +205,22 @@ struct OperationViewModel {
...
-    let optionsParams = params?.filter { param in
-        if let httpParam = param.protocol.http as? HttpParameter {
-            return allowed.contains(httpParam.in)
+private func sort(params: [ParameterType]?) -> (method: [ParameterType], optional: [ParameterType]) {
+    var methodParams = [ParameterType]()
+    var optionalParams = [ParameterType]()
+    for param in params ?? [] {
+        // TODO: Virtual Params here!!!
+        if case let .virtual(virt) = param {
+            print("\(virt.name) is Virtual Parameter for \(virt.originalParameter.name)")
```
sacheu, line 214: change print() to logger here

tjprescott: This line will go away entirely once the PR is ready :)

src/AutorestSwift/Models/Complex/ParameterType.swift
tjprescott: It seems so. While it does look like they are doing the same thing, they actually aren't.

```Swift
        switch self {
        case let .regular(param):
            // encodes a Parameter
            try container.encode(param)
        case let .virtual(param):
            // encodes a VirtualParamter
            try container.encode(param)
        }
```

I tried to just `container.encode(self)` and it caused and infinite loop and stack overflow. This is basically analogous to the logic we used elsewhere of trying to decode as one type then the next, etc.

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
```diff
@@ -166,6 +166,7 @@ class AutorestPlugin {
             SharedLogger.fail("Unable to retrieve code model from Autorest.")
         }
         let manager = Manager(withString: codeModel)
+        manager.saveCodeModel()
```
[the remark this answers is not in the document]
tjprescott: Removed.

src/AutorestSwift/Models/Complex/Parameter.swift
```diff
@@ -70,4 +70,14 @@ class Parameter: Value {
 
         try super.encode(to: encoder)
     }
+
+    public static var codingKeys: CodingKeys.Type {
+        return CodingKeys.self
+    }
```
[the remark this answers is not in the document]
tjprescott: Removed.

src/AutorestSwift/Utils/String+Extensions.swift
```diff
@@ -77,23 +77,23 @@ extension String {
...
+//        let components = splitAndJoinAcronyms
+//        let casedComponents = components.map { $0.capitalized }
+//        let value = casedComponents.joined()
+//        return value
+//    }
```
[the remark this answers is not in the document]
tjprescott: Removed.

src/AutorestSwift/Utils/String+Extensions.swift
tjprescott: Yup. It's removed.

src/AutorestSwift/Utils/String+Extensions.swift
tjprescott: Removed.

src/AutorestSwift/Code Generator/Manager.swift
tjprescott: Yeah, it's just for debugging if you want to preserve the code model file when running autorest. It won't normally be invoked. I removed the line that calls this.

tjprescott:
Resolution on where params should go. Essentially, all required params (flattened or not) go in the signature, and all optional params go in the options.

PR #168, opened 2020-08-27
Fix date issues in chat preview 2 swagger

### LINKED ISSUE #166: Date must conform to LosslessStringConvertible when in query string
Resulting error due to change in Swagger.

sacheu:
* Add more format for dateformatter to decode date in http response
* Add util function to format Date to string before putting that in query parameter
Fixes #166

### DIFF
--- a/src/AutorestSwift/Code Generator/SwiftGenerator.swift
+++ b/src/AutorestSwift/Code Generator/SwiftGenerator.swift
@@ -63,14 +63,6 @@ class SwiftGenerator: CodeGenerator {
             andParams: [:]
         )
 
-        // Create Util.swift file
-        try render(
-            template: "UtilFile",
-            toSubfolder: .util,
-            withFilename: "Util",
-            andParams: [:]
-        )
-
         // Create Enumerations.swift file
         let enumViewModel = EnumerationFileViewModel(from: model.schemas)
         try render(
--- a/src/AutorestSwift/Code Generator/SwiftGenerator.swift
+++ b/src/AutorestSwift/Code Generator/SwiftGenerator.swift
@@ -63,6 +63,14 @@ class SwiftGenerator: CodeGenerator {
             andParams: [:]
         )
 
+        // Create Util.swift file
+        try render(
+            template: "UtilFile",
+            toSubfolder: .util,
+            withFilename: "Util",
+            andParams: [:]
+        )
+
         // Create Enumerations.swift file
         let enumViewModel = EnumerationFileViewModel(from: model.schemas)
         try render(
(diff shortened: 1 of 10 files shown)


templates/UtilFile.stencil
```diff
@@ -0,0 +1,7 @@
+import Foundation
+ 
+ public func formatDateToString(_ date: Date) -> String {
+     let dateFormatter = DateFormatter()
+     dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZ"
+     return dateFormatter.string(from: date)
+ }
```
tjprescott, line 7: Does this not work?

```Swift
extension Date: LosslessStringConvertible {
    public init?(_ description: String) {
        let dateFormatter = DateFormatter()
        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZ"
        self = dateFormatters.date(from: description) ?? Date()
    }
}
```

templates/ModelDecoderDateProperty.stencil
```diff
@@ -1,7 +1,12 @@
 if let dateString = try? container.decode(String.self, forKey: .{{ property.name }}) {
     let dateFormatter = DateFormatter()
     dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSSZ"
-    self.{{ property.name }} = dateFormatter.date(from: dateString)
+    if let date = dateFormatter.date(from: dateString) {
+        self.{{ property.name }} = date
+    } else {
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZ"
+        self.{{ property.name }} = dateFormatter.date(from: dateString)
+    }
```
tjprescott, line 9: Perhaps we just has a list of date formats 
```Swift
let formats = [
  "yyyy-MM-dd'T'HH:mm:ss.SSSSZ",
  "yyyy-MM-dd'T'HH:mm:ssZ"
]
```

And then we just iterate through those until one succeeds, and then break. That way it would be easier to add other odd formats.

src/AutorestSwift/Code Generator/SwiftGenerator.swift
```diff
@@ -63,6 +63,14 @@ class SwiftGenerator: CodeGenerator {
...
         )
 
+        // Create Util.swift file
+        try render(
+            template: "UtilFile",
+            toSubfolder: .util,
+            withFilename: "Util",
+            andParams: [:]
+        )
+
```
tjprescott, line 73: So I think the filename here should be DateUtil... or, if we can make it worth with an extension, "Date+Extensions"

templates/UtilFile.stencil
tjprescott: Yeah, nevermind this LosslessString stuff. Just use AzureCore's 
`String(describing: Date, withFormat: Date.Format)` and `Date(_: String?, format: Date.Format)` methods.

PR #177, opened 2020-09-03
Update codegen for dispatch queue changes

tjprescott:
Updates necessary to support the following PR:
https://github.com/Azure/azure-sdk-for-ios/pull/415

This also updates the podfile to point to my fork for YAMS.


templates/ClientOptionsFile.stencil
```diff
@@ -15,22 +15,27 @@ public struct {{ model.name }}Options: AzureClientOptions {
     public let telemetryOptions: TelemetryOptions
     /// Global transport options
     public let transportOptions: TransportOptions
+    /// The default dispatch queue on which to call all completion handler. Defaults to `DispatchQueue.main`.
+    public let dispatchQueue: DispatchQueue?
```
sacheu, line 19: Under what situation where we should use a new queue not the `DispatchQueue.main`?

tjprescott: Your test app for one :)  Anytime you want to call a bunch of things in the background. The previous behavior of calling all completionHandlers on main would break the chain and cause the code to hang.

sacheu (approved):
Looks good. Thanks Travis.

PR #178, opened 2020-09-03
Remove casing logic

### LINKED ISSUE #153: Remove casing logic

tjprescott:
Closes #153. We can defer the casing logic to autorest and simplify our code.


sacheu (approved):
Looks good.

PR #184, opened 2020-09-10
Fix compile errors in generated testserver code in master branch

sacheu:
Filter out code generation when exception status code is `default`


AutorestSwift.xcodeproj/xcshareddata/xcschemes/AutorestSwift.xcscheme
```diff
@@ -44,7 +44,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      launchStyle = "0"
+      launchStyle = "1"
```
tjprescott, line 47: What's this for?

src/AutorestSwift/View Models/ExceptionResponseViewModel.swift
```diff
@@ -32,6 +32,7 @@ struct ExceptionResponseViewModel {
     let objectType: String?
     /// Identifies the correct snippet to use when rendering the view model
     let description: String?
+    let isDefault: Bool
```
tjprescott, line 35: This isn't very self-documenting. Is what default? Looks like it should be something like `isDefaultStatusCode`.  Is this a swagger feature where if someone doesn't set the status code for some operation it defaults to 200?

test/integration/generated/body-file/AutoRestSwaggerBatFile.xcodeproj/xcuserdata/sacheu.xcuserdatad/xcschemes/xcschememanagement.plist
```diff
@@ -12,17 +12,17 @@
 		<key>AutoRestSwaggerBatFilePackageDescription.xcscheme_^#shared#^_</key>
```
tjprescott, line 12: This file should not be committed.

src/AutorestSwift/View Models/ExceptionResponseViewModel.swift
sacheu: In Swagger file, it just say 'default' with a definition schema '#/definitions/Error'

 ```
  "responses": {                                                                                                               
          "200": {                                                                                                                   
            "description": "The PNG file",                                                                                           
            "schema": {                                                                                                              
              "type": "file"                                                                                                         
            }                                                                                                                        
          },                                                                                                                         
          "default": {                                                                                                               
            "description": "Unexpected error",                                                                                       
            "schema": {                                                                                                              
              "$ref": "#/definitions/Error"                                                                                          
            }                                                                                                                        
          }                                                                                                                          
        }     
```

File: test/integration/generated/body-file/AutoRestSwaggerBatFile.xcodeproj/xcuserdata/sacheu.xcuserdatad/xcschemes/xcschememanagement.plist (hunk 3 above)
Comment (sacheu) [reply test/integration/generated/body-file/AutoRestSwaggerBatFile.xcodeproj/xcuserdata/sacheu.xcuserdatad/xcschemes/xcschememanagement.plist]: Thanks. Reverted.

File: src/AutorestSwift/View Models/ExceptionResponseViewModel.swift (hunk 2 above)
Comment (sacheu) [reply src/AutorestSwift/View Models/ExceptionResponseViewModel.swift:35]: Yes. good point. isDefaultStatusCode is a better name.

File: templates/OperationResponseExceptionSnippet.stencil [hunk 4]
```diff
@@ -1,4 +1,5 @@
 {% for exception in op.exceptions %}
+    {% if exception.statusCodes.count > 0 %}
```
sacheu, line 2: Do not generate this code block if the exception is 'default' status code

src/AutorestSwift/View Models/ExceptionResponseViewModel.swift
```diff
@@ -32,26 +32,26 @@ struct ExceptionResponseViewModel {
...
     let description: String?
+    let isDefaultStatusCode: Bool
 
     init(from response: Response) {
         let httpResponse = response.protocol.http as? HttpResponse
         var statusCodes = [String]()
         httpResponse?.statusCodes.forEach { statusCodes.append($0.rawValue) }
 
-        self.statusCodes = statusCodes
+        self.statusCodes = statusCodes.filter { $0 != "default" }
```
tjprescott, line 42: What happens if it is default? This ends up being an empty collection? Is that the intention? What defines what the defaults are?

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -140,16 +142,28 @@ struct OperationViewModel {
...
+            }
         }
 
         var statusCodes = [String]()
         for response in responses {
             statusCodes.append(contentsOf: response.statusCodes)
         }
         for exception in exceptions {
-            statusCodes.append(contentsOf: exception.statusCodes)
+            for statusCode in exception.statusCodes where statusCode != "default" {
```
tjprescott, line 164: You already filtered out default in the initializer, no?

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
```diff
@@ -36,14 +36,26 @@ struct ReturnTypeViewModel {
...
+            for statusCode in response.statusCodes {
+                statusCodes.insert(statusCode)
+            }
+        }
+
+        assert(strategies.count <= 1, "Different strategy in ResponseViewModel is currently not supported.")
+        self.statusCodes = Array(statusCodes)
+        self.strategy = strategies.first?.rawValue ?? ResponseBodyType.noBody.rawValue
+
+        let response = responses?.first
```
tjprescott, line 53: If there are multiple responses, what's the rationale for just taking the first one?

templates/OperationResponseBodySnippet.stencil
```diff
@@ -19,7 +18,5 @@ case let .success(data):
     }
     {% include "OperationResponseExceptionSnippet.stencil" %}
 case let .failure(error):
-    DispatchQueue.main.async {
-        completionHandler(.failure(error), httpResponse)
-    }
+    {% include "OperationResponseFailureSnippet.stencil" %}
```
tjprescott, line 21: What's the difference between:
- OperationResponseFailureSnippet
- OperationResponseExceptionSnippet

The sound very very smiliar.

templates/OperationResponseExceptionSnippet.stencil
tjprescott: What is default status code? If the logic is based on the default, wouldn't you make the if statement trigger based on `isDefaultStatusCode`?

templates/OperationResponseFailureBodySnippet.stencil
```diff
@@ -0,0 +1,18 @@
+guard let data = httpResponse?.data else {
+    let noDataError = AzureError.sdk("Response data expected but not found.")
+    DispatchQueue.main.async {
+        completionHandler(.failure(noDataError), httpResponse)
+    }
```
tjprescott, line 5: This will need to be updated based on the changes to master to use the local `dispatchQueue`.

templates/OperationResponseFailureBodySnippet.stencil
```diff
@@ -0,0 +1,18 @@
...
+    return
+}
+do {
+    let decoder = JSONDecoder()
+    let decoded = try decoder.decode({{ op.defaultException.objectType }}.self, from: data)
+    DispatchQueue.main.async {
+        completionHandler(.failure(AzureError.service("{{ op.defaultException.description }}", decoded)), httpResponse)
```
tjprescott, line 12: Same here. `dispatchQueue`

templates/OperationResponseFailureBodySnippet.stencil
```diff
@@ -0,0 +1,18 @@
...
+    }
+} catch {
+     DispatchQueue.main.async {
+        completionHandler(.failure(AzureError.sdk("Decoding error.", error)), httpResponse)
```
tjprescott, line 16: `dispatchQueue`

templates/OperationResponseFailureNoBodySnippet.stencil
```diff
@@ -0,0 +1,3 @@
+DispatchQueue.main.async {
+    completionHandler(.failure(error), httpResponse)
+}
```
tjprescott, line 3: `dispatchQueue`

tjprescott (changes requested):
Needs a rebase for the DispatchQueue.main -> dispatchQueue change. But I feel like I'm missing something. We do a bunch of stuff to exclude "default" status code, but I don't see that we assertively do anything with it. So what does statusCode "default" mean to us and how should our code gen handle it?

AutorestSwift.xcodeproj/xcshareddata/xcschemes/AutorestSwift.xcscheme
sacheu: Reverted.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
sacheu: I want to fix that in a new pr instead of keep growing this one. So existing code only consider the first response, so i leave that as-is

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Good catch. Removed this change.

src/AutorestSwift/View Models/ExceptionResponseViewModel.swift
tjprescott: I just saw your explanation. It does seem then that this collection would always just be integers.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
tjprescott: Even in your multi-response PR we only consider the first one?

test/integration/generated/head/Source/AutoRestHeadTestClient.swift
```diff
@@ -254,8 +254,8 @@ public final class AutoRestHeadTestClient: PipelineClient {
                     return
                 }
                 if [
-                    204,
-                    404
+                    404,
+                    204
```
tjprescott, line 258: Perhaps we should sort the status code collection so the generated code is deterministic.

Podfile.lock
```diff
@@ -18,15 +18,15 @@ PODS:
...
+  - Yams (3.0.1)
 
 DEPENDENCIES:
   - Stencil (from `https://github.com/stencilproject/Stencil.git`, branch `trim_whitespace`)
   - SwiftFormat/CLI
   - SwiftLint
   - SwiftNIO
   - SwiftNIOFoundationCompat
-  - Yams (from `https://github.com/tjprescott/Yams.git`, branch `CircularReferences`)
+  - Yams
```
tjprescott, line 29: This looks like a regression. It should be pulling from my fork.

src/AutorestSwift/View Models/ExceptionResponseViewModel.swift
```diff
@@ -32,26 +32,26 @@ struct ExceptionResponseViewModel {
...
 
         if let errorResponseMetadata = response.extensions?["x-ms-error-response"]?.value as? Bool,
             errorResponseMetadata {
             guard objectType != nil
             else { fatalError("Did not find object type for error response") }
-        } else {
-            fatalError("Did not find object type for error response")
         }
+
+        self.isDefaultStatusCode = statusCodes.filter { $0 == "default" }.count > 0
```
tjprescott, line 55: Should just be able to do:
```Swift
self.isDefaultStatusCode = statusCodes.contains("default")
```

tjprescott: And looking at how it is implemented in the Swagger, maybe this should just be called "defaultException" since it really isn't a status code at all.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -140,8 +142,18 @@ struct OperationViewModel {
             responses.append(ResponseViewModel(from: response, with: operation))
         }
 
+        var defaultException: ExceptionResponseViewModel?
         for exception in operation.exceptions ?? [] {
-            exceptions.append(ExceptionResponseViewModel(from: exception))
+            let vm = ExceptionResponseViewModel(from: exception)
```
tjprescott, line 147: I would avoid using `vm`. In Azure-speak it means "virtual machine" and Swift frowns on variables with fewer than 3 characters anyways.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -140,8 +142,18 @@ struct OperationViewModel {
...
+            if vm.isDefaultStatusCode {
+                assert(
+                    defaultException == nil,
+                    "Multiple default exception per operation is currently not supported... \(operation.name)"
+                )
```
tjprescott, line 152: I don't think there could be multiple defaults anyhow since it is a dictionary and the keys have to be unique, right?

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
tjprescott: In that case, we should leave a TODO to come back here and fix up this logic.

templates/OperationResponseFailureSnippet.stencil
```diff
@@ -0,0 +1,5 @@
+{% if op.defaultExceptionHasBody == true %}
```
tjprescott, line 1: Maybe this file could be called "OperationDefaultExceptionSnippet" since that's what it is doing.

tjprescott (changes requested):
Some more comments now that I better understand what "default" status code is.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
sacheu: and in my multiple-exception PR, it is building code from 'exception' array in code mode, not from multiple response in the code mode.

sacheu: So we still only support 1 response (but with multiple exception) at this point.

templates/OperationResponseBodySnippet.stencil
sacheu: ResponseException is for the 'exception' array in the code mode.
ResponsFailure is when the result from the http handler returns failure.

templates/OperationResponseFailureSnippet.stencil
sacheu: Thanks. Renamed.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Renamed.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: yeah I removed the Assert.

src/AutorestSwift/View Models/ExceptionResponseViewModel.swift
sacheu: When it is default and there is a body, the handling will be generated in OperationDefaultExceptionSnippet.stencil

src/AutorestSwift/View Models/ExceptionResponseViewModel.swift
sacheu: Fixed. Thanks.

templates/OperationResponseExceptionSnippet.stencil
sacheu: if an exception is default, it will be set to the defaultException in the operationViewModel. And the values will be used by OperationDefaultExceptionSnippet.stenicl.

src/AutorestSwift/View Models/ExceptionResponseViewModel.swift
```diff
@@ -32,26 +32,26 @@ struct ExceptionResponseViewModel {
...
 
         if let errorResponseMetadata = response.extensions?["x-ms-error-response"]?.value as? Bool,
             errorResponseMetadata {
             guard objectType != nil
             else { fatalError("Did not find object type for error response") }
-        } else {
-            fatalError("Did not find object type for error response")
         }
+
+        self.defaultException = statusCodes.contains("default")
```
tjprescott, line 55: Maybe `hasDefaultException` because you have something a little later called `defaultException` and it's an `ExceptionResponsViewModel`

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -140,8 +142,14 @@ struct OperationViewModel {
             responses.append(ResponseViewModel(from: response, with: operation))
         }
 
+        var defaultException: ExceptionResponseViewModel?
         for exception in operation.exceptions ?? [] {
-            exceptions.append(ExceptionResponseViewModel(from: exception))
+            let exceptionResponseVM = ExceptionResponseViewModel(from: exception)
```
tjprescott, line 147: Would recommend just `viewModel`

tjprescott (changes requested):
Two very small nits but otherwise LGTM

tjprescott (approved):

PR #186, opened 2020-09-12
Create 2 more tests to test body.json endponts

sacheu:
* Add test for getEmptyFile and getFileLarge
* check the data length in the http response in getFile
* remove a plist file checked in by mistake
* update .gitignore


.gitignore
```diff
@@ -303,5 +303,5 @@ AutorestSwift.xcodeproj/xcuserdata/
 Pods/
 .build/
 .swiftpm/
-test/integration/generated/body-file/AutoRestSwaggerBatFile.xcodeproj/xcuserdata
-test/integration/generated/head/AutoRestHeadTest.xcodeproj/xcuserdata
+*.plist
```
tjprescott, line 306: I don't think we can just add plists to .gitignore. Some plists we need checked in because each project looks for an info.plist in its build settings.

.gitignore
```diff
@@ -303,5 +303,5 @@ AutorestSwift.xcodeproj/xcuserdata/
...
+test/integration/generated/**/**/xcuserdata
```
tjprescott, line 307: 😍  
Thank you for this!

scripts/testserver_codegen.sh
```diff
@@ -10,6 +10,5 @@ do
     autorest --input-file=./node_modules/@microsoft.azure/autorest.testserver/swagger/$i.json --output-folder=./test/integration/generated/$i --namespace=$i --use=.
     cd ./test/integration/generated/$i
     swift build
-    swift package generate-xcodeproj
     cd ../../../..
```
tjprescott, line 13: This can put you in a wonky state if the `testserver_codegen.sh` script fails. Is there any way to ensure we get back to the original starting directory?

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -157,7 +157,9 @@ struct OperationViewModel {
             statusCodes.append(contentsOf: response.statusCodes)
         }
         for exception in exceptions {
-            statusCodes.append(contentsOf: exception.statusCodes)
+            for statusCode in exception.statusCodes where statusCode != "default" {
+                statusCodes.append(statusCode)
+            }
```
tjprescott, line 162: Is `exceptions` a model object or a view model object? If it's a model object, it seems we could have a calculated property or helper method to handle this logic (I've seen it repeated, I think). If it's a view model, then I think we are already making sure that status codes is only numeric.

templates/OperationResponseFailureSnippet.stencil
```diff
@@ -0,0 +1,5 @@
+{% if op.defaultExceptionHasBody == true %}
+{% include "OperationResponseFailureBodySnippet.stencil" %}
+{% else %}
+{% include "OperationResponseFailureNoBodySnippet.stencil" %}
+{% endif %}
```
tjprescott, line 5: Wasn't this already checked in?

scripts/testserver_codegen.sh
```diff
@@ -10,6 +10,5 @@ do
     autorest --input-file=./node_modules/@microsoft.azure/autorest.testserver/swagger/$i.json --output-folder=./test/integration/generated/$i --namespace=$i --use=.
     cd ./test/integration/generated/$i
     swift build
-    swift package generate-xcodeproj
```
tjprescott, line 13: Why was this line removed?

tjprescott (changes requested):
Mostly questions.

.gitignore
sacheu: Thanks. Revert.

templates/OperationResponseFailureSnippet.stencil
sacheu: Sorry this should be removed

scripts/testserver_codegen.sh
sacheu: I need to investigate how we can do that. Will be address in other PR.

scripts/testserver_codegen.sh
sacheu: i don't want to recreate a new Xcode project for the generated code everytime because we need to modify the generated XCode project before we can add to AutorestswiftTest xcode project. And i don't want the manual changed to be overwritten everytime we run this script. The modification we need to make in the generated Xcode project is to remove the AzureCore target in each of the generated Xcode project. Without this modification, Xcode will complain saying there are multiple AzureCore targets (since each Xcode project will have its own AxureCore target by default (without the manually deletion)

tjprescott (approved):

PR #187, opened 2020-09-12
Add support for DictionarySchema in code model and Add test case for report.json

sacheu:
* Support  DictionarySchema in code model 
* Generate code for report.json testserver swagger
* xcode project changes for new test cases


sacheu:
Hand written test case is AutoRestReportTest.swift
all code under test/integration/generated/report is generated

test/integration/generated/report/AutoRestReport.xcodeproj/AutoRestReport_Info.plist
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+  <key>CFBundleDevelopmentRegion</key>
+  <string>en</string>
+  <key>CFBundleExecutable</key>
+  <string>$(EXECUTABLE_NAME)</string>
+  <key>CFBundleIdentifier</key>
```
tjprescott, line 8: So AutoRestReport is a generated library with test cases?

test/integration/generated/report/AutoRestReport.xcodeproj/xcshareddata/xcschemes/AutoRestReport-Package.xcscheme
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   LastUpgradeVersion = "9999"
+   version = "1.3">
+   <BuildAction
```
tjprescott, line 5: I think this scheme is not needed. The default `swift package generate-xcodeproj` makes a bunch of schemes that we remove. I usually remove all the schemes that that command generates and just re-add the one that, in this case, would be `AutoRestReport`.

AutorestSwiftTest/AutoRestReportTest.swift
```diff
@@ -0,0 +1,85 @@
...
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+import XCTest
+import AzureCore
+import AutoRestReport
+
+class AutoRestReportTest: XCTestCase {
```
tjprescott, line 32: So it looks like we are testing out ability to retrieve the report... but are we actually doing anything with the results?

tjprescott (changes requested):
Tweaking the auto-generated schemes. If we can do that in an automated way, that would be great.

AutorestSwiftTest/AutoRestReportTest.swift
sacheu: Yes, getting the report is the first step. Need to follow the rest of the steps from Laurent.

test/integration/generated/report/AutoRestReport.xcodeproj/AutoRestReport_Info.plist
sacheu: Yes, the test case is to retrieve the report.

AutorestSwiftTest/AutoRestReportTest.swift
tjprescott: Do those steps take place in here or are they an outside thing?

sacheu: Outside of this test case.

test/integration/generated/report/AutoRestReport.xcodeproj/xcshareddata/xcschemes/AutoRestReport-Package.xcscheme
sacheu: Removed other schemes that we don't need.

test/integration/generated/body-file/AutoRestSwaggerBatFile.xcodeproj/xcshareddata/xcschemes/AutoRestSwaggerBatFile-Package.xcscheme
```diff
@@ -1,24 +1,58 @@
...
-    selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-    shouldUseLaunchSchemeArgsEnv = "YES"
-    codeCoverageEnabled = "NO">
-    <Testables>
-    </Testables>
-  </TestAction>
+<Scheme
+   LastUpgradeVersion = "9999"
+   version = "1.3">
+   <BuildAction
```
tjprescott, line 5: This scheme should also go away. There should only be an `AutoRestSwaggerBatFile.xcscheme`

test/integration/generated/head/AutoRestHeadTest.xcodeproj/xcshareddata/xcschemes/AutoRestHeadTest-Package.xcscheme
```diff
@@ -1,24 +1,58 @@
...
-    selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-    shouldUseLaunchSchemeArgsEnv = "YES"
-    codeCoverageEnabled = "NO">
-    <Testables>
-    </Testables>
-  </TestAction>
+<Scheme
+   LastUpgradeVersion = "9999"
+   version = "1.3">
+   <BuildAction
```
tjprescott, line 5: Same.

test/integration/generated/report/AutoRestReport.xcodeproj/xcshareddata/xcschemes/AutoRestReport-Package.xcscheme
tjprescott: This one should be removed as well. There should only be an `AutoRestReport.xcscheme` which `generate-xcodeproj` doesn't create. That's why you remove all the ones it does create and then add this one. I can show you via screenshare if you want.

tjprescott (changes requested):
The schemes still look off to me.

AutorestSwiftTest/AutoRestReportTest.swift
sacheu: Actually, I was wrong. i look at the test report in Python side, it `print_report` after it gets the report https://github.com/Azure/autorest.python/blob/autorestv3/test/vanilla/AcceptanceTests/test_zzz.py.

tjprescott (approved):

PR #189, opened 2020-09-13
Add support for multiple responses

sacheu:
Uss testsever xms-error-responses.json as an example add support for multiple responses 
* Change decode logic to build responses, exceptions array in Operation to contain both SchemaRespons and Reponse in the array. (Before it has to be either all of them are SchemaResponse or all of them are Reponse
* Refactor OperationResponse stencil to generate code for each of the response and exception in the Operation (instead of just consider the first one)
* Move name logic of ReturnTypeViewModel to ResponseViewModel
* Create test case for xms-error-responses.json to test with test server
* Add generated code for xms-error-responses.json and other  testserver json due to this change.


src/AutorestSwift/View Models/ExceptionResponseViewModel.swift
```diff
@@ -52,6 +53,15 @@ struct ExceptionResponseViewModel {
...
 
+        switch objectType {
+        case "String":
+            self.strategy = ResponseBodyType.stringBody.rawValue
+        case "Int":
+            self.strategy = ResponseBodyType.intBody.rawValue
+        default:
+            self.strategy = ResponseBodyType.jsonBody.rawValue
+        }
+
```
tjprescott, line 64: Exception response can be int or string? Is this a testserver thing?

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -90,8 +90,7 @@ struct OperationViewModel {
     let returnType: ReturnTypeViewModel?
     let params: OperationParameters
     let pipelineContext: [KeyValueViewModel]?
-    private let requests: [RequestViewModel]?
```
tjprescott, line 93: What happened to this?

src/AutorestSwift/View Models/ResponseViewModel.swift
```diff
@@ -33,14 +33,17 @@ enum ResponseBodyType: String {
     case body
     /// Service returns no response data, only a status code
     case noBody
+    case stringBody
+    case intBody
+    case jsonBody
```
tjprescott, line 38: What is the difference between `body` and the three other body types?

src/AutorestSwift/View Models/ResponseViewModel.swift
```diff
@@ -53,20 +56,33 @@ struct ResponseViewModel {
...
             self.pagedElementClassName = nil
+            if let objectType = schemaResponse?.schema.swiftType(optional: false) {
+                strategy = .body
+                self.objectType = objectType
+            } else {
+                strategy = .noBody
+                self.objectType = "Void"
+            }
         }
+        self.strategy = strategy.rawValue
```
tjprescott, line 86: I don't see any logic setting `intBody`, `stringBody` or `jsonBody`.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
```diff
@@ -31,31 +31,14 @@ import Foundation
...
-
-        let response = responses?.first
-        self.pagingNames = response?.pagingNames
-        if let elementType = response?.pagedElementClassName, response?.pagingNames != nil {
-            self.name = "PagedCollection<\(elementType)>"
-        } else {
-            self.name = response?.objectType ?? "Void"
-        }
+        let objectTypes = Array(objectType)
+        self.name = objectTypes.count > 1 ? "Any" : objectTypes.first ?? "Void"
```
tjprescott, line 42: Do we have any examples of something returning type `Any`?  Is this a testserver test? I can't imagine in a real-world example this being a good Swift UX.

templates/OperationResponseExceptionSnippet.stencil
```diff
@@ -1,14 +1,20 @@
 {% for exception in op.exceptions %}
-    {% if exception.hasDefaultException == false %}
```
tjprescott, line 2: Is this a regression? What if there is no defaultException?

templates/OperationResponseExceptionSnippet.stencil
```diff
@@ -1,14 +1,20 @@
...
         do {
-            let decoder = JSONDecoder()
-            let decoded = try decoder.decode({{ exception.objectType }}.self, from: data)
+            {% if exception.strategy == "intBody" %}
+                {% include "OperationResponseExceptionIntBodySnippet.stencil" %}
+            {% elif exception.strategy == "stringBody" %}
+                {% include "OperationResponseExceptionStringBodySnippet.stencil" %}
+            {% elif exception.strategy == "jsonBody" %}
+                {% include "OperationResponseExceptionJsonBodySnippet.stencil" %}
+            {% else %}
```
tjprescott, line 15: These snippets are so small I'd be okay with them being inline here. `OperationResponseExceptionIntBodySnippet` doesn't aid my comprehension, particularly if this is the only place it's used.

tjprescott: We can aid the readability and maintainability of this by using stencil comments
`{# ... #} for comments not included in the template output`

tjprescott (commented):
Questions and comments.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: we don't need to keep a property of requests array in ObjectionViewmodel since we only consider the first request anywayre.

src/AutorestSwift/View Models/ExceptionResponseViewModel.swift
sacheu: Yes. that is an example in testserver xms-error-responses.json

src/AutorestSwift/View Models/ResponseViewModel.swift
sacheu: it is set in ./View Models/ExceptionResponseViewModel.swift

templates/OperationResponseExceptionSnippet.stencil
sacheu: this is not a regression. this is a necessary if check in the stencil. Because we already filter out default Exception in  op.exceptions when we build the view model. i.e. all exceptions in op.op.exceptions will be hasDefaultException equals to false. No point to have another check

src/AutorestSwift/View Models/ResponseViewModel.swift
sacheu: these 3 body types just dedicate how we decode the message body. 
* jsonBody - use JsonDecoder
* string - decode with String.init
* int - decode with String.init and then covert to int

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
sacheu: The xms-error-responses.json returns `Any`. For 202 status code, it returns nothing, but for 200 , it returns Pet object.

image

tjprescott: Perhaps then, it makes more sense to return `Pet?`

src/AutorestSwift/Code Generator/SwiftGenerator.swift
```diff
@@ -62,6 +62,13 @@ class SwiftGenerator: CodeGenerator {
             withFilename: "PatchUtil",
             andParams: [:]
         )
+        // Create StringInt+ExtensionFile.swift file
+        try render(
+            template: "StringInt+ExtensionFile",
+            toSubfolder: .util,
+            withFilename: "StringInt+ExtensionFile",
+            andParams: [:]
+        )
```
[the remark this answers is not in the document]
sacheu: Rename to Primitives +ExtensionFile.swift

src/AutorestSwift/View Models/ResponseViewModel.swift
tjprescott: But where do we actually determine the right one and set it? i only see in the init where we choose between body, pagedBody and noBody, which implies to me that we are overloading this param in a way we shouldn't.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
tjprescott: The scenario would need to be a non-error response of `.noBody` and exactly one `.body` response. Otherwise, I think we should throw an assertion failure.

src/AutorestSwift/View Models/ResponseViewModel.swift
sacheu: It is set in ExceptionResponseViewModel 
image

templates/OperationResponseExceptionSnippet.stencil
sacheu: Fixed.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
sacheu: Thanks for the suggestion. Changed.

templates/OperationResponseNoBodySnippet.stencil
```diff
@@ -1,16 +1,7 @@
...
-    {% include "OperationResponseStatusCodeSnippet.stencil" %}
-    {% include "OperationResponseReturnTypeStatusCodeSnippet.stencil" %}
-    {
-        dispatchQueue.async {
-            completionHandler(.success(()), httpResponse)
-        }
+{% include "OperationResponseStatusCodeSnippet.stencil" %}
+{
+    dispatchQueue.async {
+        completionHandler(.success({% if op.returnType.returnNil  %} nil {% else %} () {% endif %})
```
sacheu, line 4: I tried to put this into a prototype of RetrunTypeViewModel, sometthing like
let returnObject = isReturnTypeNilable ? "nil" : "()"
and then in stencil file do
completoinHandler(.success( {{ op.returnType.ReturnObject }})) , httpResponse)

But stencil engine can't resolved `op.returnType.ReturnObject` as long as it has value 'nil'.  So I do above to work around this stencil engine issue.

src/AutorestSwift/View Models/ResponseViewModel.swift
tjprescott: Ah, then I think it would make more sense, instead of adding those values to ResponseBodyType, to make a new enum ExceptionResponseBodyType, that has these values. Otherwise, one enum is describing two different things, which could make future maintenance more difficult.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
```diff
@@ -31,31 +31,32 @@ import Foundation
...
+        // Only supprt at most 1 body or pageBody return type now.
+        let withBodyStrategies = strategies.filter { $0 != "noBody" }
+        let hasNoBodyStratgies = strategies.contains { $0 == "noBody" }
+        assert(withBodyStrategies.count <= 1)
 
-        let response = responses?.first
-        self.pagingNames = response?.pagingNames
-        if let elementType = response?.pagedElementClassName, response?.pagingNames != nil {
-            self.name = "PagedCollection<\(elementType)>"
+        // since we only support 1 body/pagePage response, only need to take the first item fro objectTypes
```
tjprescott, line 53: nit: type fro -> for

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
```diff
@@ -31,31 +31,32 @@ import Foundation
...
+        if let name = objectTypes.first {
+            self.name = hasNoBodyStratgies ? "\(name)?" : name
```
tjprescott, line 55: typo: hasNoBodyStratgies -> hasNoBodyStrategies

templates/OperationResponseNoBodySnippet.stencil
tjprescott: Couldn't you bind it to a string property that's set in the view model as "nil" or "()"?

test/integration/generated/xms-error-responses/Source/XmsErrorResponseExtensionsClient.swift
```diff
@@ -0,0 +1,318 @@
...
+
+public final class XmsErrorResponseExtensionsClient: PipelineClient {
+    /// API version of the  to invoke. Defaults to the latest.
+    public enum ApiVersion: String {
+        /// API version ""
+        case v = ""
+
+        /// The most recent API version of the
+        public static var latest: ApiVersion {
+            return .v
```
tjprescott, line 27: Is this a bug in the code gen?  I guess as long as it works.

templates/OperationResponseNoBodySnippet.stencil
sacheu: That was the approach i took , create a property in RetrunTypeViewModel with name returnObject and the value of returnObject is either 'nil' or '();. But when render Stencil file stencil , it can't resolve that property for some reason

tjprescott: I don't know why Github isn't showing my reply here, but couldn't you have a string property on the view model that is either "nil" or "()"?

test/integration/generated/xms-error-responses/Source/XmsErrorResponseExtensionsClient.swift
sacheu: i think that is because the test swagger does not have api version. I think we can improve code gen by  adding this version block conditionally

src/AutorestSwift/View Models/ResponseViewModel.swift
tjprescott: Github swallowed my comment, so I'll repost. :)  

It seems this should be a different enum, `ExceptionResponseBodyStrategy` instead of being folded into `ResponseBodyStrategy` since the two sets of values are mutually exclusive. For maintainability, combining them could be confusing (i.e. the difference between .body and .bodyJson?)

templates/OperationResponseNoBodySnippet.stencil
tjprescott: Per our discussion, let's leave it as but put a comment in the Stencil file. That way we can revisit it in the future.

test/integration/generated/xms-error-responses/Source/XmsErrorResponseExtensionsClient.swift
tjprescott: I'm okay with adding that to this PR or opening an issue to do that as future work.

tjprescott:
Also, this has a conflict that needs to be resolved.

src/AutorestSwift/View Models/ResponseViewModel.swift
sacheu: Good point. Thanks.

test/integration/generated/xms-error-responses/Source/XmsErrorResponseExtensionsClient.swift
sacheu: I create https://github.com/Azure/autorest.swift/issues/192

src/AutorestSwift/View Models/ResponseViewModel.swift
tjprescott, line 38: Since these are now in ExceptionResponseBodyType, do they still need to be here?

tjprescott (commented):
One last comment.

tjprescott (approved):

PR #191, opened 2020-09-14
Use type in CodeModel to determine which Schema to be used during dec…

sacheu:
…oded

### DIFF
--- a/src/AutorestSwift/Models/Complex/Property.swift
+++ b/src/AutorestSwift/Models/Complex/Property.swift
@@ -42,7 +42,7 @@ class Property: Value {
     let isDiscriminator: Bool?
 
     enum CodingKeys: String, CodingKey {
-        case readOnly, serializedName, flattenedNames, isDiscriminator // , schema
+        case readOnly, serializedName, flattenedNames, isDiscriminator
     }
 
     // MARK: Codable
--- a/src/AutorestSwift/Models/Complex/Property.swift
+++ b/src/AutorestSwift/Models/Complex/Property.swift
@@ -42,7 +42,7 @@ class Property: Value {
     let isDiscriminator: Bool?
 
     enum CodingKeys: String, CodingKey {
-        case readOnly, serializedName, flattenedNames, isDiscriminator, schema
+        case readOnly, serializedName, flattenedNames, isDiscriminator // , schema
     }
 
     // MARK: Codable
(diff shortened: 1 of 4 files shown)


src/AutorestSwift/Models/Complex/Property.swift
```diff
@@ -42,7 +42,7 @@ class Property: Value {
     let isDiscriminator: Bool?
 
     enum CodingKeys: String, CodingKey {
-        case readOnly, serializedName, flattenedNames, isDiscriminator, schema
+        case readOnly, serializedName, flattenedNames, isDiscriminator // , schema
```
tjprescott, line 45: Why are we commenting out schema?

src/AutorestSwift/Models/Complex/Value.swift
```diff
@@ -74,7 +74,7 @@ class Value: Codable, LanguageShortcut {
 
     enum CodingKeys: String, CodingKey {
         case schema
-        case internalRequired = "required"
+        case required // = "required"
```
tjprescott, line 77: We should probably just remove the comment if it isn't needed.

sacheu:
We can't can't decode DictionarySchema and ArraySchema consistently. Because both  DictionarySchema and ArraySchema have the same required properties. if we decode DictionarySchema before ArraySchema , our private swagger failed to generate code. if we change the order, report.json failed to generate code. The solution is to look up 'type' in the code model to decide which schema to decode.

src/AutorestSwift/Models/Complex/SchemaResponse.swift
```diff
@@ -43,15 +43,22 @@ class SchemaResponse: Response {
...
+        switch type {
+        case .array:
+            schema = try? container.decode(ArraySchema.self, forKey: .schema)
+        case .dictionary:
+            schema = try? container.decode(DictionarySchema.self, forKey: .schema)
+        case .object:
+            schema = try? container.decode(ObjectSchema.self, forKey: .schema)
+        default:
+            schema = try container.decode(Schema.self, forKey: .schema)
         }
```
tjprescott, line 59: Could we wrap this logic into a Schema initializer? That would make it more reusable. 
`init(withType type: AllSchemaTypes, andContainer container: KeyedContainer)`

src/AutorestSwift/Models/Complex/Value.swift
```diff
@@ -87,4 +87,67 @@ class Value: Codable, LanguageShortcut {
...
+            schema = try? container.decode(NumberSchema.self, forKey: .schema)
+        case .choice:
+            schema = try? container.decode(ChoiceSchema.self, forKey: .schema)
+        case .dateTime:
+            schema = try? container.decode(DateTimeSchema.self, forKey: .schema)
+        case .string:
+            schema = try? container.decode(StringSchema.self, forKey: .schema)
+        default:
+            schema = try container.decode(Schema.self, forKey: .schema)
+        }
```
tjprescott, line 117: This is why an initializer would be nice. Consolidate all of _this_

tjprescott (changes requested):

src/AutorestSwift/Models/Complex/Property.swift
sacheu: Removed.

src/AutorestSwift/Models/Complex/Value.swift
sacheu: Removed.

src/AutorestSwift/Models/Complex/SchemaResponse.swift
sacheu: Create a common generic function to decode Schema. It needs to be generic because we need to pass the CodingKeys of different container class which has `schema` as its property.

src/AutorestSwift/Models/Complex/Value.swift
sacheu: Found a solution to be able to consolidate all Schema creation.

tjprescott (approved):

PR #195, opened 2020-09-15
Body Parameter Flattening

tjprescott:
Adds support for `VirtualParameter`s and `x-ms-client-flattening` for the method body parameter.

Also introduces a significant refactoring of the `OperationViewModel` class. 

For the generated code example, see: https://github.com/Azure/azure-sdk-for-ios-pr/pull/26

### DIFF
--- a/src/AutorestSwift/Models/Complex/CodeModel.swift
+++ b/src/AutorestSwift/Models/Complex/CodeModel.swift
@@ -31,7 +31,7 @@ class CodeModel: Codable, LanguageShortcut {
     let info: Info
     let schemas: Schemas
     let operationGroups: [OperationGroup]
-    let globalParameters: [Parameter]?
+    let globalParameters: [ParameterType]?
     let security: Security
     var language: Languages
     let `protocol`: Protocols
@@ -43,12 +43,12 @@ class CodeModel: Codable, LanguageShortcut {
     }
 
     /// Lookup a global parameter by name.
-    func globalParameter(for name: String) -> Parameter? {
-        return globalParameters?.first { $0.name == name }
+    func globalParameter(for name: String) -> ParameterType? {
+        return globalParameters?.first(named: name)
     }
 
     func getApiVersion() -> String {
-        if let constantSchema = globalParameter(for: "apiVersion")?.schema as? ConstantSchema {
+        if let constantSchema = globalParameter(for: "api-version")?.schema as? ConstantSchema {
             return constantSchema.value.value
         } else {
             return ""
--- a/src/AutorestSwift/Models/Complex/Parameter.swift
+++ b/src/AutorestSwift/Models/Complex/Parameter.swift
@@ -67,3 +67,11 @@ class Parameter: Value, CustomDebugStringConvertible {
         return debugString() ?? description
     }
 }
+
+extension Parameter: Equatable {
+    static func == (lhs: Parameter, rhs: Parameter) -> Bool {
+        // TODO: Improve this? This is technically the same assumption being
+        // made when using a dictionary.
+        return lhs.name == rhs.name
+    }
+}
--- a/src/AutorestSwift/Models/Complex/Parameter.swift
+++ b/src/AutorestSwift/Models/Complex/Parameter.swift
@@ -27,7 +27,7 @@
 import Foundation
 
 /// A definition of an discrete input for an operation
-class Parameter: Value {
+class Parameter: Value, CustomDebugStringConvertible {
     /// suggested implementation location for this parameter
     let implementation: ImplementationLocation?
 
@@ -59,5 +59,11 @@ class Parameter: Value {
         if flattened != nil { try container.encode(flattened, forKey: .flattened) }
 
         try super.encode(to: encoder)
+    }
+
+    // MARK: CustomDebugStringConvertible
+
+    public var debugDescription: String {
+        return debugString() ?? description
     }
 }
--- a/src/AutorestSwift/Models/Complex/ParameterType.swift
+++ b/src/AutorestSwift/Models/Complex/ParameterType.swift
@@ -1,0 +1,118 @@
+// --------------------------------------------------------------------------
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+//
+// The MIT License (MIT)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the ""Software""), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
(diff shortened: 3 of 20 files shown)


src/AutorestSwift/Models/Complex/Request.swift
```diff
@@ -54,24 +54,40 @@ class Request: Metadata {
...
     }
 }
 
 extension Request {
     /// Retrieves a single body-encoded parameter, if there is one. Fails if there is more than one.
-    var bodyParam: Parameter? {
-        var bodyParams = [Parameter]()
+    var bodyParam: ParameterType? {
+        // using a dictionary allows us to eliminate duplicates in a rudimentary fashion. A Set<T> would be
+        // better, but requires model class to conform to Hashable
```
tjprescott, line 66: While this is true, both Dictionary and Set have the side effect of turning an ordered list into an unordered one. I am currently working on a fix for this, as it means the generated code is not deterministic.

src/AutorestSwift/Models/Complex/CodeModel.swift
```diff
@@ -43,12 +43,12 @@ class CodeModel: Codable, LanguageShortcut {
...
     /// Lookup a global parameter by name.
-    func globalParameter(for name: String) -> Parameter? {
-        return globalParameters?.first { $0.name == name }
+    func globalParameter(for name: String) -> ParameterType? {
+        return globalParameters?.first(named: name)
     }
 
     func getApiVersion() -> String {
-        if let constantSchema = globalParameter(for: "apiVersion")?.schema as? ConstantSchema {
+        if let constantSchema = globalParameter(for: "api-version")?.schema as? ConstantSchema {
```
sacheu, line 51: the '-' is added to 'api-versoin'. Will that break anything?

src/AutorestSwift/Models/Complex/Parameter.swift
```diff
@@ -60,4 +60,18 @@ class Parameter: Value {
...
+
+    // MARK: CustomDebugStringConvertible
+
+    public var debugDescription: String {
+        return debugString() ?? description
+    }
+}
+
+extension Parameter: Equatable {
+    static func == (lhs: Parameter, rhs: Parameter) -> Bool {
```
sacheu, line 72: Question: When do we need to == Parameter?

src/AutorestSwift/Models/Complex/ParameterType.swift
```diff
@@ -0,0 +1,212 @@
...
+        case let .regular(lparam):
+            if case let ParameterType.regular(rparam) = rhs {
+                return lparam == rparam
+            }
+        case let .virtual(lparam):
+            if case let ParameterType.virtual(rparam) = rhs {
+                return lparam == rparam
+            }
+        }
+        return false
```
sacheu, line 119: Maybe move this "return false" to default: of the switch block?

src/AutorestSwift/Models/Complex/ParameterType.swift
```diff
@@ -0,0 +1,212 @@
...
+                return httpParam.in == .query
+            } else {
+                return false
+            }
+        }
+    }
+
+    /// Returns the items that should be passed in the request header
+    var inHeader: [ParameterType] {
+        return filter { param in
```
sacheu, line 187: the method inHeader, InPath, inSignature have. a commonitity. May be create a helper function and take what to check in httpParam.in for each function?

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -30,19 +30,17 @@ struct OperationParameters {
...
-        parameters: [Parameter]?,
-        operation: Operation,
-        operationParameters: OperationParameters? = nil
-    ) {
-        self.header = operationParameters?.header ?? Params()
-        self.query = operationParameters?.query ?? Params()
-        self.path = operationParameters?.path ?? [KeyValueViewModel]()
-
-        for param in parameters ?? [] {
+    init(parameters: [ParameterType], operation: Operation) {
```
sacheu, line 39: Why we don't need to pass in operationParameters anymore?

src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -35,23 +35,25 @@ enum RequestBodyType: String {
     case patchBody
 }
 
-struct BodyParam {
+struct BodyParamViewModel {
```
sacheu, line 38: Should this be in its own file?

sacheu (changes requested):
Looks good. Just some questions/suggestions.

src/AutorestSwift/Models/Complex/CodeModel.swift
tjprescott: Without it, it wasn't generating the API names correctly. They were showing up just a "v".

src/AutorestSwift/Models/Complex/Parameter.swift
tjprescott: To use `params.contains(param)` it requires that `ParameterType` conform to `Equatable`, and thus `Parameter` must as well.

tjprescott: Otherwise I was just using a dictionary to simulate the uniqueness, but that destroyed the ordering (as would using `Set<T>`)

src/AutorestSwift/Models/Complex/ParameterType.swift
tjprescott: When I try that, I get a warning that the `default will never be executed` and that it is `missing return in a function expected to return 'Bool'`

src/AutorestSwift/Models/Complex/ParameterType.swift
tjprescott: Done.

src/AutorestSwift/View Models/OperationViewModel.swift
tjprescott: `Operation.allParams` collects any parameters from the Request, which is why we used to have this. We would collect the params from the operation and then go back through and make a copy, updating with the params from the Request.

src/AutorestSwift/View Models/RequestViewModel.swift
tjprescott: Yes, it should. Done.

src/AutorestSwift/Models/Complex/CodeModel.swift
tjprescott: So, according to David:
```
Modelerfour normalizes api version parameters for you, you shouldn't have to do anything special.  In this case, they either need to rename it in the swagger to api-version or use x-ms-api-version: true
```

sacheu (approved):

PR #196, opened 2020-09-16
Suppot non json body in response 

sacheu:
1. Add support for 3 different body in response:
* json
* int
* string
2. refactor how code generation for 3 different body works in exception response
3. Add support for different Int type: Int32 Int64
4. Add nullable response suppport


AutorestSwiftTest/AutoRestBodyIntegerTest.swift
```diff
@@ -0,0 +1,83 @@
...
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import XCTest
+import AzureCore
+import AutoRestIntegerTest
+
+class AutoRestIntegerTest: XCTestCase {
```
sacheu, line 31: This test case is stil work in progress.

src/AutorestSwift/Utils/ViewModelUtil.swift
```diff
@@ -0,0 +1,39 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+import Stencil
```
tjprescott, line 29: Does this actually use Stencil or Foundation?

src/AutorestSwift/Utils/ViewModelUtil.swift
```diff
@@ -0,0 +1,39 @@
...
+
+public func getStrategy(_ objectType: String) -> String {
+    if objectType == "String" {
+        return ResponseBodyType.stringBody.rawValue
+    } else if objectType.starts(with: "Int") {
+        return ResponseBodyType.intBody.rawValue
+    } else {
+        return ResponseBodyType.jsonBody.rawValue
+    }
+}
```
tjprescott, line 39: This seems like it should be a function on the enum type. 

```swift
enum ResponseBodyType {
  func strategy(for input: String) -> ResponseBodyType  {
    switch input {
      case "String":
        return .stringBody
      // etc
    }
  }

}
```

src/AutorestSwift/View Models/ResponseViewModel.swift
```diff
@@ -73,13 +78,12 @@ struct ResponseViewModel {
             self.pagingNames = nil
             self.pagedElementClassName = nil
             if let objectType = schemaResponse?.schema.swiftType(optional: false) {
-                strategy = .body
+                self.strategy = getStrategy(objectType)
```
tjprescott, line 81: I'm a little curious why we are alternating between `strategy` and `self.strategy`. Aren't we passing the code through `swiftlint autocorrect` and `swiftformat`?

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
```diff
@@ -38,21 +38,25 @@ struct ReturnTypeViewModel {
     init(from responses: [ResponseViewModel]?) {
         var objectTypes = Set<String>()
         var strategies = Set<String>()
+        var isNullables = Set<Bool>()
```
tjprescott, line 41: What is the purpose of this? If you are trying to align objectType, strategy and nullability using an index, I don't think that works.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
```diff
@@ -38,21 +38,25 @@ struct ReturnTypeViewModel {
...
+            isNullables.insert(response.isNullable)
         }
 
         // Only supprt at most 1 body or pageBody return type now.
         let withBodyStrategies = strategies.filter { $0 != "noBody" }
-        let hasNoBodyStrategies = strategies.contains { $0 == "noBody" }
         assert(withBodyStrategies.count <= 1)
 
+        let hasNoBodyStrategies = strategies.contains { $0 == "noBody" }
+        let hasNullableResponse = isNullables.contains { $0 == true }
```
tjprescott, line 55: Hmm, why not something like:

```Swift
var hasNullableResponse = false
for response in responses ?? [] {
  hasNullableResponse = hasNullableResponse || response.isNullable
}
```
Or
```Swift
let hasNullableResponse = responses.first(where: { $0.isNullable == true }) != nil
```

tjprescott: If `strategies` isn't used elsewhere except to set the value of `hasNoBodyStrategies` then I'd recommend a similar approach for that too.

templates/OperationResponseNullableBodyEndSnippet.stencil
```diff
@@ -0,0 +1,3 @@
+{% if response.isNullable == true %}
+}
+{% endif %}
```
tjprescott, line 3: I'm not sure what this is for. NullableBodyEndSnippet? Seems like this should live inline.

src/AutorestSwift/View Models/ResponseViewModel.swift
sacheu: i think it is possible that i build using 'make install' (instead of XCode) more now. And the 'swiftlint autocorrect and swiftformat' are setup in Xcode workspace.'

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
sacheu: strategies Set is used to check in the assert to check if there is more than 1 body/pagedBody.  We can remove that once we can remove that assert. So leaving that as is for now.

src/AutorestSwift/View Models/ReturnTypeViewModel.swift
sacheu: Thanks. Removed.

templates/OperationExceptionIntBodySnippet.stencil
```diff
@@ -0,0 +1,12 @@
...
+    let decoded = {{ exception.objectType }}(decodedstr) {
+        dispatchQueue.async {
+        completionHandler(.failure(AzureError.service("{{ exception.description }}", decoded)), httpResponse)
+    }
+} else {
+    dispatchQueue.async {
+        completionHandler(.failure(AzureError.sdk("Decoding error.", nil)), httpResponse)
+    }
+}
+{% include "OperationExceptionNullableBodyEndSnippet.stencil" %}
```
tjprescott, line 12: Is this EndSnippet still a thing?

templates/OperationExceptionStringBodySnippet.stencil
```diff
@@ -0,0 +1,6 @@
+{% include "OperationExceptionNullableBodyStartSnippet.stencil" %}
+let decoded = String(data: data, encoding: .utf8)
+dispatchQueue.async {
+    completionHandler(.failure(AzureError.service("{{ exception.description }}", decoded)), httpResponse)
+}
+{% include "OperationExceptionNullableBodyEndSnippet.stencil" %}
```
tjprescott, line 6: Same here

src/AutorestSwift/View Models/ResponseViewModel.swift
tjprescott: We should definitely have `make install` run it through the same process so as not to introduce random back-and-forth changes that aren't really part of the PR.

src/AutorestSwift/Utils/ViewModelUtil.swift
sacheu: Thanks. Changed per suggestion.

src/AutorestSwift/Utils/ViewModelUtil.swift
sacheu: File is removed.

src/AutorestSwift/View Models/ResponseViewModel.swift
sacheu: yeah. I will address this in a separate pr.

templates/OperationExceptionIntBodySnippet.stencil
sacheu: Thanks. Removed.

templates/OperationResponseNullableBodyEndSnippet.stencil
sacheu: File is removed.

tjprescott (approved):

PR #200, opened 2020-09-23
Fix operation.exceptions to be optional

tjprescott:
Fixes code generation failure when swagger doesn't contain exceptions.


sacheu (approved):
Thanks Travis.

PR #201, opened 2020-09-23
Bug fixes for code generate another swagger

sacheu:


AutorestSwift.xcodeproj/project.pbxproj
```diff
@@ -301,6 +303,8 @@
 			dstPath = "";
 			dstSubfolderSpec = 7;
 			files = (
+				F198F81C251BC456009CB4E5 /* OperationResponseStatusCodeDeclareSnippet.stencil in CopyFiles */,
+				F198F81D251BC456009CB4E5 /* OperationResponseUnixTImeBodySnippet.stencil in CopyFiles */,
```
tjprescott, line 307: Typo: OperationResponseUnixT**I**meBodySnippet -> OperationResponseUnixTimeBodySnippet

src/AutorestSwift/Code Generator/Manager.swift
```diff
@@ -157,6 +157,10 @@ class Manager {
                     .warn(
                         "Discrepancies found in round-tripped code model. Run a diff on 'before.json' and 'after.json' to troubleshoot."
                     )
+                SharedLogger
+                    .warn("beforeJsonUrl=\(beforeJsonUrl)")
+                SharedLogger
+                    .warn("afterJsonUrl=\(afterJsonUrl)")
```
tjprescott, line 163: Was this just present for debugging?

src/AutorestSwift/View Models/ParameterViewModel.swift
```diff
@@ -36,8 +36,9 @@ struct ParameterViewModel {
     let defaultValue: ViewModelDefault
     let comment: ViewModelComment
 
-    init(from param: ParameterType, withName name: String? = nil) {
-        self.name = name ?? param.name
+    init(from param: ParameterType, withName specificName: String? = nil) {
+        let name = specificName ?? param.name
+        self.name = name.isEmpty ? param.name : name
```
tjprescott, line 41: This is a little longer, but seems easier to understand for me. Also happy to leave as is.
```suggestion
        if let name = specificName, !name.isEmpty {
          self.name = name
        } else {
          self.name = param.name
        }
```

Though it is odd that someone would pass in a specific name of empty string. Is this common or would this be a swagger bug?

tjprescott (changes requested):
Some small comments.

AutorestSwift.xcodeproj/project.pbxproj
sacheu: Thanks. Renamed

src/AutorestSwift/Code Generator/Manager.swift
sacheu: I leave that thinking it can be useful for future debugging. But I have reverted this change.

src/AutorestSwift/View Models/ParameterViewModel.swift
sacheu: I have applied ^ suggestion.

src/AutorestSwift/Code Generator/Manager.swift
tjprescott: It saved the two files to the generated folder, which would be easier to compare since you can diff with FileMerge (or something similar). My concern leaving this in would be just log spam.

tjprescott (approved):

PR #203, opened 2020-09-24
Add functionalities in autorest swift code generate to support test swagger url.json 

sacheu:
1. Split up operation group in different files if the operation group has a name
2. Split up the Option files in this naming convention {OperationName}.{OptionFile}.swift
3. If there is a name collision between Operation file name and Model file name, append 'Operation' to the operation File name.
4. For schema with implementation == 'method' , add snippet to decode datatype into String so that it can be used to construct query param and header and path. Handle these cases:
     i) the value is a constant
    ii) the value is from signature of the function
    iii) the value is from option of the method

5. For schema with implementation == 'client' add property in the generated Client code.
6. For required client property, add the parameter in the init() of the generated Client code.


sacheu:
All generated code for these swagger
"head" "body-file" "report" "xms-error-responses" "body-integer" "url"
 compiled

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -49,45 +61,123 @@ struct KeyValueViewModel {
...
             self.value = ""
             self.optional = false
-            self.paramName = nil
+            self.constantValue = nil
+            self.needDecodingInMethod = false
         }
+
+        self.key = name
+        self.strategy = keyValueType.rawValue
     }
```
tjprescott, line 140: This seems like a very large initializer. It might make more sense to split this into different convenience intializers.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -49,45 +61,123 @@ struct KeyValueViewModel {
...
+        return "\(key ?? value)String"
+    case .choice,
+         .sealedChoice:
+        return "\(value).rawValue"
+    case .array:
+        return "\(value).map { String($0) }.joined(separator: \",\") "
+    default:
+        return "\(value)"
     }
 }
```
tjprescott, line 183: I think this would be unnecessary if `ConstantValue.value` where declared type `Any`

src/AutorestSwift/View Models/OperationGroupViewModel.swift
```diff
@@ -31,7 +31,7 @@ import Foundation
 ///     // MARK: OperationGroupName
 ///     ...
 struct OperationGroupViewModel {
-    let name: String
+    var name: String
```
tjprescott, line 34: Why the switch to var?

src/AutorestSwift/View Models/ParameterViewModel.swift
```diff
@@ -38,9 +38,9 @@ struct ParameterViewModel {
 
     init(from param: ParameterType, withName specificName: String? = nil) {
         if let name = specificName, !name.isEmpty {
-            self.name = name
+            self.name = name.firstLowerCased
         } else {
-            self.name = param.name
+            self.name = param.name.firstLowerCased
         }
```
tjprescott, line 44: Can we not control this in our autorest.swift configuration? We eliminated most if not all casing logic in our code by using what autorest configuration already provides.

src/AutorestSwift/View Models/PropertyViewModel.swift
```diff
@@ -34,15 +34,17 @@ struct PropertyViewModel {
     let name: String
     let comment: ViewModelComment
     let type: String
+    // default value of the proeprty
     let defaultValue: ViewModelDefault
+    // default value of the property in the init(). Valid values are either nil (for optoinal property) or "" (i.e. not specfied for required property in init() method.)
     let initDefaultValue: String
```
tjprescott, line 40: There's a typo in the second comment (optoinal), but I'm not sure what is the rationale here? If there's a default value, why not just put that in the init?

src/AutorestSwift/View Models/PropertyViewModel.swift
```diff
@@ -34,15 +34,17 @@ struct PropertyViewModel {
...
     let isDate: Bool
     let optional: Bool
     let className: String
 
     /// Initialize from Value type (such as Property or Parameter)
     init(from schema: Value) {
-        self.name = schema.name
+        self.name = schema.serializedName ?? schema.name
```
tjprescott, line 47: Would it make sense to have an assertion that the name is not empty string? That way we are alerted to the problem during code gen instead of when we are looking at the generated code?

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
```diff
@@ -35,18 +35,46 @@ struct ServiceClientFileViewModel {
     let apiVersionName: String
     let protocols: String
     let paging: Language.PagingNames?
+    let globalParameters: [ParameterViewModel]
+    // A dictionary of all the named operation group. Key is the group name , Value is the operation group view model.
+    let namedOperationGroups: [String: OperationGroupViewModel]
+    // A key,Value pairs of all the named operation group for stencil template engine
+    let namedOperationGroupProperties: [KeyValueViewModel]
```
tjprescott, line 42: Wouldn't these live inside the `OperationGroupViewModel`? Why are they here?

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
```diff
@@ -35,18 +35,46 @@ struct ServiceClientFileViewModel {
...
         self.name = "\(model.packageName)Client"
         self.comment = ViewModelComment(from: model.description)
-        var items = [OperationGroupViewModel]()
+        var operationGroups = [OperationGroupViewModel]()
+        var namedOperationGroups = [String: OperationGroupViewModel]()
+        var namedOperationGroupProperties = [KeyValueViewModel]()
         for group in model.operationGroups {
-            items.append(OperationGroupViewModel(from: group, with: model))
+            var viewModel = OperationGroupViewModel(from: group, with: model)
+            var groupName = group.name.trimmingCharacters(in: .whitespacesAndNewlines)
```
tjprescott, line 52: We could build this normalizing logic into the getter so that any time you get `group.name` you always get the trimmed version.

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
```diff
@@ -35,18 +35,46 @@ struct ServiceClientFileViewModel {
...
+            if groupName.isEmpty {
+                operationGroups.append(viewModel)
+                // if the operation group name has a collision with a model object, append 'Operation' to the operation group name
+            } else if model.object(for: groupName) != nil {
+                groupName += "Operation"
+                viewModel.name = groupName
+                namedOperationGroups[groupName] = viewModel
+            } else {
+                namedOperationGroups[groupName] = viewModel
+            }
```
tjprescott, line 62: It seems like the initializer for `OperationGroupViewModel` should resolve the name conflict, instead of doing it here.

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
```diff
@@ -35,18 +35,46 @@ struct ServiceClientFileViewModel {
...
+        self.operationGroups = operationGroups
+        self.namedOperationGroups = namedOperationGroups
         self.apiVersion = model.getApiVersion()
         self.apiVersionName = "v\(apiVersion.replacingOccurrences(of: "-", with: ""))"
         self.paging = model.pagingNames
         self.protocols = paging != nil ? "PipelineClient, PageableClient" : "PipelineClient"
+        var globalParameters = [ParameterViewModel]()
+        for globalParameter in model.globalParameters ?? [] where !globalParameter.name.starts(with: "$") {
+            globalParameters.append(ParameterViewModel(from: globalParameter))
+        }
```
tjprescott, line 73: What happens with global parameters that do start with $?

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
```diff
@@ -35,18 +35,46 @@ struct ServiceClientFileViewModel {
...
+        self.globalParameters = globalParameters
+        for key in namedOperationGroups.keys {
+            namedOperationGroupProperties.append(KeyValueViewModel(key: key, value: key.lowercased()))
+        }
+        self.namedOperationGroupProperties = namedOperationGroupProperties
```
tjprescott, line 78: Is there a reason this appears after global parameters? If so, a comment would be helpful. If not, it might be easier from a maintainability perspective to put this after line 65.

tjprescott: Though I'm still not clear why this wouldn't just be located in the OperationGroupViewModel

tjprescott (changes requested):
I didn't look very closely at the stencil files and commented mostly on the view model logic.

src/AutorestSwift/Models/Complex/ConstantSchema.swift
```diff
@@ -54,7 +58,28 @@ class ConstantSchema: Schema {
...
+                    try valueContainer.encode(Double(value.value), forKey: .value)
+                } else {
+                    try valueContainer.encode(value.value, forKey: .value)
+                }
+            } else {
+                try valueContainer.encode(Double(value.value), forKey: .value)
+            }
+        default:
+            try valueContainer.encode(value.value, forKey: .value)
+        }
```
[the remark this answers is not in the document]
sacheu: `This logic should live in the encode method for ConstantValue.swift. `. I can't put this logic in ConstantValue.swift since it does not have the `type` information there. The logic is in ConstantSchema , since it has both `type` and `vaule`

src/AutorestSwift/Models/Complex/Schema.swift
```diff
@@ -84,7 +84,8 @@ class Schema: Codable, LanguageShortcut {
     func swiftType(optional: Bool = false) -> String {
         var swiftType: String
         switch type {
-        case AllSchemaTypes.string:
+        case AllSchemaTypes.string,
+             AllSchemaTypes.constant:
```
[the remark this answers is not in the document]
sacheu: I will take your implementation instead

src/AutorestSwift/Models/Complex/NumberSchema.swift
```diff
@@ -86,11 +86,11 @@ class NumberSchema: PrimitiveSchema {
...
-            return "Int32"
+            return (type == AllSchemaTypes.integer) ? "Int32" : "Float"
         case 64:
-            return "Int64"
+            return (type == AllSchemaTypes.integer) ? "Int64" : "Double"
         default:
-            return "Int"
+            return (type == AllSchemaTypes.integer) ? "Int" : "Float"
         }
     }
```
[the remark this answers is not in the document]
sacheu: Both Schema and NumberSchema does not have property indicating whether it is optional or not. 
It is the schema using Schema and NumberSchema to show whether they are optional.

src/AutorestSwift/View Models/ClientMethodOptionsViewModel.swift
```diff
@@ -40,14 +40,23 @@ struct ClientMethodOptionsViewModel {
     // The list of properites for the options object
     let properties: [PropertyViewModel]
 
-    init(from operation: Operation, with model: CodeModel, parameters: [ParameterType]) {
+    let extensionName: String
+
+    init(from operation: Operation, with model: CodeModel, parameters: [ParameterType], groupName: String) {
```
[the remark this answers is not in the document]
sacheu: The group name is taken from `Language` 's `name` property. And that is String, not String?. If there is no name, it will be an empty string.

src/AutorestSwift/View Models/ClientMethodOptionsViewModel.swift
```diff
@@ -40,14 +40,23 @@ struct ClientMethodOptionsViewModel {
...
         self.clientName = model.name
         self.operationName = operation.name
         self.name = "\(operation.name)Options"
+        if model.object(for: groupName) != nil {
+            self.extensionName = groupName + "Operation"
+        } else {
+            self.extensionName = groupName
+        }
```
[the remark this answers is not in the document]
sacheu: Yes. I have added this comment `  // if the operation group name has a collision with a model object, append 'Operation' to the operation group name`

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
sacheu: Removed. Thanks.

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
sacheu: Changed per feedback.

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
```diff
@@ -36,13 +36,17 @@ struct ServiceClientFileViewModel {
     let protocols: String
     let paging: Language.PagingNames?
     let globalParameters: [ParameterViewModel]
+    // A dictionary of all the named operation group. Key is the group name , Value is the operation group view model.
     let namedOperationGroups: [String: OperationGroupViewModel]
```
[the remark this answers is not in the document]
tjprescott: Since this isn't iterated through in the codegen (only to generate the files), this shouldn't be an issue.

src/AutorestSwift/Models/Complex/ConstantSchema.swift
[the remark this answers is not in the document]
tjprescott: Interesting. Then perhaps ConstantValue should be moved as an internal struct of ConstantSchema to ensure it can't be instantiated by itself.

src/AutorestSwift/Models/Complex/NumberSchema.swift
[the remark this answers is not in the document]
tjprescott: But even so, if that schema does say it is optional, does it not need to be plumbed into this method so it comes back as `Int64?`?

src/AutorestSwift/Models/Complex/ConstantSchema.swift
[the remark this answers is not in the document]
sacheu: Move ConstantValue to inside Constant.

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
sacheu: Rename to namedOperationGroupShortcuts

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
sacheu: This is for creating properties for each named operation group in ServiceClient.

src/AutorestSwift/Models/Complex/Schema.swift
```diff
@@ -100,14 +101,18 @@ class Schema: Codable, LanguageShortcut {
             } else {
                 swiftType = "[\(name)]"
             }
-        case AllSchemaTypes.dateTime:
+        case AllSchemaTypes.dateTime,
+             AllSchemaTypes.date:
             swiftType = "Date"
+        case AllSchemaTypes.byteArray:
+            swiftType = "[UInt8]"
```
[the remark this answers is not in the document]
sacheu: Changed to Date

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Broke up into small init() for each scenario

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -153,31 +153,31 @@ struct KeyValueViewModel {
...
-         .sealedChoice:
-        return "\(value).rawValue"
-    case .array:
-        return "\(value).map { String($0) }.joined(separator: \",\") "
-    default:
-        return "\(value)"
+    /**
+     Convert the type into String format in Swift
+     */
+    private static func formatValueForType(type: AllSchemaTypes, value: String, key: String? = nil) -> String {
```
tjprescott, line 160: The double declaration of type should be avoided. The "Swifty" way:
`formatValue(forType type: ...)`

templates/OperationOptionsDecodingSnippet.stencil
```diff
@@ -1,6 +1,6 @@
-{% if param.strategy == "dateFromConstant" %}
+{% if param.strategy == "dateFromParam" %}
 let {{ param.key }}String = DateFormatter().string(from: {{ param.key }})
-{% elif param.strategy == "byteArrayFromConstant" %}
+{% elif param.strategy == "byteArrayFromParam" %}
```
tjprescott, line 3: Was this name changed because the param may not necessarily be a constant?

tjprescott (changes requested):
A couple new comments and a few unaddressed ones. Overall looks great :)

templates/OperationOptionsDecodingSnippet.stencil
sacheu: Yes. I can be from one of the parametrs in the Option.

src/AutorestSwift/Models/Simple/Schemas.swift
```diff
@@ -47,7 +47,7 @@ class Schemas: Codable {
     let strings: [StringSchema]?
 
     /// UnixTime
-    let unixTimes: [UnixTimeSchema]?
+    let unixtimes: [UnixTimeSchema]?
```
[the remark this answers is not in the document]
sacheu: Fixed by adding CodingKeys

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Rename the function to clarify the purpose of this function.

src/AutorestSwift/View Models/ParameterViewModel.swift
sacheu: Revert this change.

tjprescott (approved):

PR #204, opened 2020-09-28
Update script to publish to public repo.

tjprescott:


scripts/git_sync.sh
```diff
@@ -42,9 +42,9 @@ echo "Switching to: " $SDK_REPO_ROOT
 cd $SDK_REPO_ROOT
 
 echo
-echo "Checking out preview/SecretSDK branch"
+echo "Checking out communication/ChatService branch"
```
sacheu, line 45: should we use 'feature/communication' branch instead?

tjprescott:
Updated

sacheu (approved):

PR #209, opened 2020-10-06
Add 10 new test cases from url.json

sacheu:
* Add 10 new test cases from url.json (5 from paths, 5 from queries)
* Fix autorest swift code generator in order to get these test cases to work

    1.  Add base64 encoding in path/query parameters 
    2. For the yams workaround, we need to leave the '' for 'value' in code model unchanged (i.e. do not replace `value: '' ` to ` value: ' ' `


src/AutorestSwift/Code Generator/Manager.swift
```diff
@@ -81,7 +81,11 @@ class Manager {
     private static func sanitize(yaml: String) -> String {
         // TODO: Remove when issue (https://github.com/Azure/autorest.swift/issues/47) is fixed.
         // Replaces empty string with a single space.
-        return yaml.replacingOccurrences(of: ": ''", with: ": ' '")
+        // Keep the empty string for Value. Replace all empty string for value with PreserveEmptyString. After globally replace
+        // '' with ' ', replace PreserveEmptyString with ''
+        let preserveEmptyValue = yaml.replacingOccurrences(of: "value: ''", with: "value: PreserveEmptyString")
+        let removedGlobalEmpty = preserveEmptyValue.replacingOccurrences(of: ": ''", with: ": ' '")
+        return removedGlobalEmpty.replacingOccurrences(of: "value: PreserveEmptyString", with: "value: ''")
```
tjprescott, line 88: What is the rationale for this?

templates/OperationOptionsDecodingSnippet.stencil
```diff
@@ -1,7 +1,10 @@
 {% if param.strategy == "dateFromParam" %}
 let {{ param.key }}String = DateFormatter().string(from: {{ param.key }})
 {% elif param.strategy == "byteArrayFromParam" %}
-let {{ param.key }}String = String(bytes: {{ param.key }}, encoding: .utf8)
+guard let {{ param.key }}String = String(bytes: {{ param.key }}, encoding: .utf8)?.data(using:.utf8)?.base64EncodedString() else {
+    self.options.logger.error("Failed to construct String for {{ param.key }}")
+    return
+}
```
tjprescott, line 7: Why are we converting to base64EncodedString?

tjprescott (commented):
Looks good. Couple questions.

src/AutorestSwift/Code Generator/Manager.swift
sacheu: In yaml file of url.json, it has ConstantSchema with '' as value:

```
  - !<!ConstantSchema> &ref_27                                                                                        
      type: constant                                                                                                    
      value: !<!ConstantValue>                                                                                          
        value: ''                                                                                                       
      valueType: *ref_0                                                                                                 
      language: !<!Languages>                                                                                           
        default:                                                                                                        
          name: Constant14                                                                                              
          description: ''             
```  
And this constantSchema is used as path/query parameter in one of the generated method.
But if we globally replacing all '' to ' ' in the yaml file, the value of this constant will changed incorrectly by ^ workaround. As a result, we build the path/query incorrectly in the generated code. So we need to skip changing '' to ' ' in `value:`

sacheu: Revert the change. Open this issue to track for a better solution for the yams issue. https://github.com/Azure/autorest.swift/issues/214

templates/OperationOptionsDecodingSnippet.stencil
sacheu: Looking at generated code in other langs. In most langs, base64 encode is done in the generated code. So I have reverted this change and do the base64 encoded in the test case itself. Test cases pass with the new change.

tjprescott (approved):

PR #211, opened 2020-10-06
model-flattening.json swagger support

tjprescott:
This PR introduced the necessary modifications needed to make `model-flattening.json` generate and compile. A follow-up PR will be created to actually introduce the tests in order to keep this PR from growing.

Work to do:
- [x] Create aggregate model file for grouped parameters
- [x] Create aggregate models with parent relations
- [x] Populate params with paths from parameter groupings
- [x] Eliminate stored properties that recursively contain itself
- [x] Generate model constants


scripts/testserver_codegen.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
```
tjprescott, line 1: These changes will need to be reverted once the PR is finalized.

scripts/testserver_codegen.sh
```diff
@@ -10,7 +11,8 @@ else
     exit
 fi
 
-files=( "head" "body-file" "report" "xms-error-responses" "body-integer" "url")
+# files=( "head" "body-file" "report" "xms-error-responses" "body-integer" "url" "model-flattening" )
+files=( "model-flattening" )
```
tjprescott, line 15: Needs to be fixed prior to final PR.

scripts/testserver_codegen.sh
```diff
@@ -21,7 +23,7 @@ do
 	rm -Rf ./test/integration/generated/$i/.build
     fi
     echo "autorest --input-file=./node_modules/@microsoft.azure/autorest.testserver/swagger/$i.json --output-folder=./test/integration/generated/$i --namespace=$i --use=."
-    autorest --input-file=./node_modules/@microsoft.azure/autorest.testserver/swagger/$i.json --output-folder=./test/integration/generated/$i --namespace=$i --use=. > /dev/null 2>&1
+    autorest --input-file=./node_modules/@microsoft.azure/autorest.testserver/swagger/$i.json --output-folder=./test/integration/generated/$i --namespace=$i --use=. --debug
```
tjprescott, line 26: Fix prior to final PR.

tjprescott:
```Swift
/// Put Flattened Simple Product with client flattening true on the parameter
/// - Parameters:
///    - flattenedSimpleProduct : Simple body product to post
///    - productId : Unique identifier representing a specific product for a given latitude & longitude. For example, uberX in San Francisco will have a different product_id than uberX in Los Angeles.
///    - options: A list of options for the operation
///    - completionHandler: A completion handler that receives a status code on
///     success.
public func postFlattenedSimpleProduct(
    productId: String,
    withOptions options: PostFlattenedSimpleProductOptions? = nil,
    completionHandler: @escaping HTTPResultHandler<SimpleProduct>
) {
    ...
    // Construct request
    let body = SimpleProduct(
        productId: productId,
        description: options?.description,
        maxProductDisplayName: options?.maxProductDisplayName,
        genericValue: options?.genericValue,
        odataValue: options?.odataValue,
        capacity: options?.capacity
    )
    ...
}
```

Right now this is the status of the flattened test swagger. Issues with this code:

- [ ] The body param appears in the doc string when it doesn't in the signature.
- [ ] The flattened options don't appear in the docstring or signature
- [ ] The flattened values are being pulled from the options object, not the signature

This should look like this
```Swift
/// Put Flattened Simple Product with client flattening true on the parameter
/// - Parameters:
///    - productId : Unique identifier representing a specific product for a given latitude & longitude. For example, uberX in San Francisco will have a different product_id than uberX in Los Angeles.
///    - description: Description
///    ...
///    - options: A list of options for the operation
///    - completionHandler: A completion handler that receives a status code on
///     success.
public func postFlattenedSimpleProduct(
    productId: String,
    description: String? = nil,
    ...
    withOptions options: PostFlattenedSimpleProductOptions? = nil,
    completionHandler: @escaping HTTPResultHandler<SimpleProduct>
) {
    ...
    // Construct request
    let body = SimpleProduct(
        productId: productId,
        description: description,
        maxProductDisplayName: maxProductDisplayName,
        genericValue: genericValue,
        odataValue: odataValue,
        capacity: capacity
    )
    ...
}
```

AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        client = try AutoRestResourceFlatteningTestClient(
+            baseUrl: baseUrl,
+            authPolicy: AnonymousAccessPolicy(),
+            withOptions: AutoRestResourceFlatteningTestClientOptions()
+        )
+    }
+
+//    func test_Head_success200() throws {
```
tjprescott, line 47: I will make a separate PR to introduce these tests. This PR is already quite large.

src/AutorestSwift/Code Generator/SwiftGenerator.swift
```diff
@@ -43,6 +43,72 @@ class SwiftGenerator: CodeGenerator {
...
+        // Skips regenerating a file with a less specific schema (Schema vs ObjectSchema, for example)
+        // but keeps track of the number of times the model name appears for debugging purposes.
+        var modelsWritten = [String: Int]()
+
+        // Create model files
+        for object in model.schemas.objects ?? [] {
+            let name = object.name
+            guard modelsWritten[name] == nil else {
+                SharedLogger.warn("\(name) has already been generated once this run. Skipping...")
+                modelsWritten[name]! += 1
```
sacheu, line 57: how does  modelsWritten[name] int Value being used? From what I can see, it sees like we can use Bool instead of Int in this lookup dictionary

src/AutorestSwift/Models/Complex/ConstantSchema.swift
```diff
@@ -76,10 +76,8 @@ class ConstantSchema: Schema {
 
     public required init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
-
-        valueType = try Schema.decode(withContainer: container, useKey: "valueType") ?? container
-            .decode(Schema.self, forKey: .valueType)
-        value = try container.decode(ConstantValue.self, forKey: .value)
+        valueType = try Schema.decode(withContainer: container, useKey: "valueType")!
```
sacheu, line 79: why we don't try to decode '
            container..decode(Schema.self, forKey: .valueType)'  when  Schema.decode failed? I think it is possible that we don't have all the possible Schema handled in Schema.decode.

src/AutorestSwift/Models/Complex/Value.swift
```diff
@@ -29,7 +29,7 @@ import Foundation
 /// Common base interface for properties, parameters and the like.
 class Value: Codable, LanguageShortcut {
     /// the schema of this Value
-    public var schema: Schema
+    public var schema: Schema?
```
sacheu, line 32: Shouldn't Value always have a schema? Why `schema` changes to optional here?

src/AutorestSwift/Models/Complex/ArraySchema.swift
```diff
@@ -28,7 +28,7 @@ import Foundation
 
 class ArraySchema: Schema {
     /// elementType of the array
-    let elementType: Schema
+    let elementType: Schema?
```
sacheu, line 31: Shouldn't ArraySchema always have a elementType? Why `elementType` changes to optional here?

src/AutorestSwift/Models/Complex/Parameter.swift
```diff
@@ -66,6 +74,28 @@ class Parameter: Value, CustomDebugStringConvertible {
     public var debugDescription: String {
         return debugString() ?? description
     }
+
+    internal func belongsInSignature() -> Bool {
+        // We track body params in a special way, so always omit them generally from the signature.
+        // If they belong in the signature, they will be added in a way to ensure they come first.
+        guard paramLocation != .body else { return false }
+
+        // Default logic
```
sacheu, line 83: What does 'default logicl' here means?

src/AutorestSwift/Code Generator/SwiftGenerator.swift
tjprescott: That's true, but it doesn't actually matter what we use. We only look to see if the value is nil, meaning there hasn't been an entry. I originally used `Bool` but switched to `Int` since, for debugging purposes, it might be useful to see how many times a file would have otherwise been regenerated. For example, if you see that one file has a much larger count than others, it might point to a name conflict.

src/AutorestSwift/Models/Complex/ConstantSchema.swift
tjprescott: Because we already have that as the default case in `Schema.decode(withContainer:useKey:)`. We could put it in the catch clause as a fallback, but it doesn't see that this behavior was needed by anything in Autorest.

src/AutorestSwift/Models/Complex/Value.swift
tjprescott: It should, and yet, in decoding `model-flattening.json` it turned out there were many "required" properties that weren't supplied. I'll try reverting the change and see if it plays nicely now. It's possible those errors were the result of something else.

src/AutorestSwift/Models/Complex/ArraySchema.swift
tjprescott: Same reason as Schema in the Value class.

sacheu (changes requested):
Looks good. Just a few questions.

src/AutorestSwift/Models/Complex/Parameter.swift
tjprescott: The special casing logic appears above, so this represents the default logic. I also changed it to make it hopefully more self-documenting, and split it between Parameter and VirtualParameter since they have different logic.

src/AutorestSwift/View Models/ObjectViewModel.swift
```diff
@@ -33,20 +33,79 @@ import Foundation
 struct ObjectViewModel {
     let name: String
     let comment: ViewModelComment
-    let objectType = "struct"
     let properties: [PropertyViewModel]
-    let isErrorType: Bool
+    let constants: [ConstantViewModel]
+    let hasConstants: Bool
+
+    var inheritance = "NSObject"
```
sacheu, line 40: Why we need to inherit from "NSObject", can this not specify?

tjprescott: Right, I debated what to put here. In this case NSObject is a legitimate value, but would essentially be a sentinel that something went wrong since it should be "Codable" with, possible, "Swift.Error". I spent a lot of time trying to figure out what was happening with a struct that was showing only "Codable" because it was the default. A sentinel would have saved a lot of troubleshooting.

src/AutorestSwift/Models/Complex/ArraySchema.swift
tjprescott: I was able to revert this back to being required without issue.

src/AutorestSwift/Models/Complex/Value.swift
tjprescott: I was able to revert this back to required without issue.

src/AutorestSwift/Models/Complex/ConstantSchema.swift
tjprescott: I updated the catch clause of `Schema.decode` to try this instead of just automatically returning nil.

tjprescott:
@sacheu thanks for the review. I pushed a new iteration based on your comments.

src/AutorestSwift/Models/Complex/Schema.swift
```diff
@@ -135,25 +136,27 @@ class Schema: Codable, LanguageShortcut {
...
-                schema = try? container.decode(ChoiceSchema.self, forKey: keyEnum)
+                schema = try container.decode(ChoiceSchema.self, forKey: keyEnum)
             case .dateTime:
-                schema = try? container.decode(DateTimeSchema.self, forKey: keyEnum)
+                schema = try container.decode(DateTimeSchema.self, forKey: keyEnum)
             case .string:
-                schema = try? container.decode(StringSchema.self, forKey: keyEnum)
+                schema = try container.decode(StringSchema.self, forKey: keyEnum)
             case .constant:
                 schema = try? container.decode(ConstantSchema.self, forKey: keyEnum)
```
sacheu, line 161: should this be 'try' instead of 'try?' too?

tjprescott: No, instead it should be the remaining cases with `try?` should be `try`, but great catch! Thank you!

PR #213, opened 2020-10-06
Remove failedExpectation in all test cases

sacheu:
Remove failedExpectation in all test cases. Call XCTFail when result is not expected. 
This will allow the test case to fail faster, instead of waiting for it to timeout.


AutorestSwiftTest/XmsErrorResponseExtensionsTest.swift
```diff
@@ -163,24 +149,36 @@ class XmsErrorResponseExtensionsTest: XCTestCase {
...
+            return
+        }
+        print(testString)
+
+        guard let test2String = String(bytes: test, encoding: .utf8)?.data(using: .utf8)?.base64EncodedString() else {
+            return
+        }
+        print(test2String)
+    }
+
```
tjprescott, line 174: Remove?

tjprescott (changes requested):
Looks great. Looks like some debug code remains though.

AutorestSwiftTest/XmsErrorResponseExtensionsTest.swift
sacheu: Yes. Removed.

tjprescott (approved):

PR #215, opened 2020-10-07
Add skipUrlEncoding in generated code

sacheu:
Change Autorest Swift to 
* read baseUrl from code model
* skip url encoding flag from code model

### DIFF
--- a/src/AutorestSwift/Models/Complex/Value.swift
+++ b/src/AutorestSwift/Models/Complex/Value.swift
@@ -128,4 +128,12 @@ class Value: Codable, LanguageShortcut {
         try container.encode(`protocol`, forKey: .protocol)
         if extensions != nil { try container.encode(extensions, forKey: .extensions) }
     }
+
+    var isSkipUrlEncoding: Bool {
+        var skipUrlEncoding: Bool?
+        if let value = extensions?["x-ms-skip-url-encoding"] {
+            skipUrlEncoding = value.value as? Bool
+        }
+        return skipUrlEncoding ?? false
+    }
 }
--- a/src/AutorestSwift/Models/Complex/Value.swift
+++ b/src/AutorestSwift/Models/Complex/Value.swift
@@ -130,10 +130,10 @@ class Value: Codable, LanguageShortcut {
     }
 
     var isSkipUrlEncoding: Bool {
-        var skipUrlEncoding: Bool?
-        if let value = extensions?["x-ms-skip-url-encoding"] {
-            skipUrlEncoding = value.value as? Bool
+        var skipUrlEncoding = false
+        if let value = extensions?["x-ms-skip-url-encoding"]?.value as? Bool {
+            skipUrlEncoding = value
         }
-        return skipUrlEncoding ?? false
+        return skipUrlEncoding
     }
 }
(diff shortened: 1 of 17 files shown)


src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -82,8 +82,18 @@ struct KeyValueViewModel {
...
         if type == .string {
-            self.value = "\"\(constantValue)\""
-            self.constantValue = "\"\(constantValue)\""
+            var skipUrlEncoding: Bool?
+            var value: String
+            if let value = param.value.extensions?["x-ms-skip-url-encoding"] {
+                skipUrlEncoding = value.value as? Bool
+            }
+            if skipUrlEncoding ?? false {
+                value = "\"\(constantValue)\".removingPercentEncoding ?? \"\(constantValue)\""
```
tjprescott, line 91: I would expect that if `x-ms-skip-url-encoding` is true, then using the right-hand side as a fallback would fail. Perhaps this should just fail, or empty string?

tjprescott (commented):
LGTM. Couple suggestions.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -82,8 +82,18 @@ struct KeyValueViewModel {
         let type = constantSchema.valueType.type
 
         if type == .string {
-            self.value = "\"\(constantValue)\""
-            self.constantValue = "\"\(constantValue)\""
+            var skipUrlEncoding: Bool?
+            var value: String
+            if let value = param.value.extensions?["x-ms-skip-url-encoding"] {
+                skipUrlEncoding = value.value as? Bool
+            }
```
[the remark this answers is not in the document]
sacheu: Good suggestion. Change per feedback.

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Yeah. Change to empty string.

src/AutorestSwift/Models/Complex/Value.swift
```diff
@@ -128,4 +128,12 @@ class Value: Codable, LanguageShortcut {
         try container.encode(`protocol`, forKey: .protocol)
         if extensions != nil { try container.encode(extensions, forKey: .extensions) }
     }
+
+    var isSkipUrlEncoding: Bool {
+        var skipUrlEncoding: Bool?
+        if let value = extensions?["x-ms-skip-url-encoding"] {
+            skipUrlEncoding = value.value as? Bool
+        }
+        return skipUrlEncoding ?? false
```
tjprescott, line 137:
```suggestion
        return extensions?["x-ms-skip-url-encoding"]?.value as? Bool ?? false 
```

tjprescott (commented):
One suggestion, but LGTM.

src/AutorestSwift/Models/Complex/Value.swift
tjprescott: Less concise, but still a simplification:
```suggestion
        var skipUrlEncoding = false
        if let value = extensions?["x-ms-skip-url-encoding"]?.value as? Bool {
            skipUrlEncoding = value
        }
        return skipUrlEncoding
```

tjprescott (approved):

PR #217, opened 2020-10-08
Makefile updates

tjprescott:
- Adds cleanAll to clean the root and test folders. 
- Keeps clean for just cleaning the root folder. 
- Updates make build to also rebuild AzureCore for use in tests following a cleanAll.


sacheu (approved):
Looks good. Thanks for fixing this.

PR #218, opened 2020-10-09
Add date datetime test cases in url.json

sacheu:


src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -28,10 +28,12 @@ import Foundation
 
 enum KeyValueDecodeStrategy: String {
     case dateFromParam
-    case byteArrayFromParam
-    case `default`
     case dateFromSignature
```
tjprescott, line 31: Why no dataFromParam?

sacheu: There is dateFromParam, line 30. I just move down byteArrayFromParam to line 32.

tjprescott (approved):

PR #221, opened 2020-10-13
Model Flattening Tests

tjprescott:
Adds 11 tests for the `model-flattening.json` swagger.

- [x] Make sure optional arguments have nil defaults in signatures and initializers
- [x] GroupSchema does not correctly deserialize GroupProperties. Duplicates appear in generated group model schemas. (Solution likely similar to Parameter/VirtualParameter)

4 tests pass. The remaining 7 tests will be worked in follow-up PRs and are tracked in #159.

### DIFF
--- a/AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
+++ b/AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
@@ -268,13 +268,11 @@ class AutoRestResourceFlatteningTest: XCTestCase {
             )
         let group = FlattenParameterGroup(
             name: "name",
-            simpleBodyProduct: SimpleProduct(
-                productId: "Product1",
-                description: "Description"
-            )
+            productId: "Product1",
+            description: "Description"
         )
         client.autorestresourceflatteningtestservice
-            .putSimpleProduct(withGrouping: group) { result, httpResponse in
+            .putSimpleProductWithGrouping(flattenParameterGroup: group) { result, httpResponse in
                 switch result {
                 case .success:
                     expectation.fulfill()
--- a/AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
+++ b/AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
@@ -44,21 +44,304 @@ class AutoRestResourceFlatteningTest: XCTestCase {
         )
     }
 
-//    func test_Head_success200() throws {
-//        let expectation = XCTestExpectation(description: "Call head200 succeed")
-//        let failedExpectation = XCTestExpectation(description: "Call head200 failed")
-//        failedExpectation.isInverted = true
-//
-//        client.head200() { result, _  in
-//            switch result {
-//                case .success:
-//                 expectation.fulfill()
-//               case let .failure(error):
-//                print("test failed. error=\(error.message)")
-//                failedExpectation.fulfill()
-//            }
-//        }
-//
-//        wait(for: [expectation], timeout: 5.0)
-//    }
+    private func errorDetails(for error: AzureError, withResponse response: HTTPResponse?) -> String {
+        var details: String
+        if let data = response?.data {
+            details = String(data: data, encoding: .utf8)!
+        } else {
+            details = error.message
+        }
+        return details
+    }
+
+    func test_resourceFlattening_getArray() throws {
+        let expectation = XCTestExpectation(description: "Call autorestresourceflatteningtestservice.getArray succeed")
+
+        client.autorestresourceflatteningtestservice.getArray { result, httpResponse in
+            switch result {
+            case .success:
+                expectation.fulfill()
+            case let .failure(error):
+                let details = self.errorDetails(for: error, withResponse: httpResponse)
+                print("test failed. error=\(details)")
+                XCTFail("Call autorestresourceflatteningtestservice.getArray failed")
+            }
+        }
+        wait(for: [expectation], timeout: 5.0)
+    }
+
+    func test_resourceFlattening_putArray() throws {
+        let expectation =
+            XCTestExpectation(
+                description: "Call autorestresourceflatteningtestservice.putArray succeed"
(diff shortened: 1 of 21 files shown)


AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
```diff
@@ -44,21 +44,304 @@ class AutoRestResourceFlatteningTest: XCTestCase {
...
+    }
+
+    func test_resourceFlattening_putArray() throws {
+        let expectation =
+            XCTestExpectation(
+                description: "Call autorestresourceflatteningtestservice.putArray succeed"
+            )
+        let array = [
+            Resource(id: nil, type: nil, tags: ["tag1": "value1", "tag2": "value3"], location: "West US", name: nil),
+            Resource(id: nil, type: nil, tags: nil, location: "Building 44", name: nil)
```
tjprescott, line 80: @sacheu it seems like this initializer should have `= nil` defaults for its optional arguments to simplify. Do you agree?

AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
```diff
@@ -44,21 +44,304 @@ class AutoRestResourceFlatteningTest: XCTestCase {
...
+                XCTFail("Call autorestresourceflatteningtestservice.putDictionary failed")
+            }
+        }
+        wait(for: [expectation], timeout: 5.0)
+    }
+
+    func test_resourceFlattening_getResourceCollection() throws {
+        let expectation =
+            XCTestExpectation(description: "Call autorestresourceflatteningtestservice.getResourceCollection succeed")
+        client.autorestresourceflatteningtestservice.getResourceCollection { result, httpResponse in
```
tjprescott, line 156: @sacheu is there a more concise way of getting to the methods? This seems very verbose, but may just be a result of the swagger.

AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
```diff
@@ -44,21 +44,304 @@ class AutoRestResourceFlatteningTest: XCTestCase {
...
+                odataValue: nil,
+                productId: "Product1",
+                description: "Description"
+            ),
+            productId: "Product1",
+            description: nil,
+            maxProductDisplayName: nil,
+            genericValue: nil,
+            odataValue: "Description"
+        )
```
tjprescott, line 307: This looks like a code generation error. The parameters are in the grouped parameter but still are in the signature.

AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
```diff
@@ -44,21 +44,304 @@ class AutoRestResourceFlatteningTest: XCTestCase {
...
+        client.autorestresourceflatteningtestservice
+            .putSimpleProductWithGrouping(flattenParameterGroup: group) { result, httpResponse in
```
tjprescott, line 309: This seems like a failure of the naming heuristics. I would think this signature should look like:
`client.blah.putSimpleProduct(withGrouping: group)`

@sacheu, what do you think?

tjprescott (commented):
@sacheu would love your take on the overall structure of the tests. Pointed out a few areas I'd like your specific thoughts on as well.

AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
tjprescott: I have fixed this.

AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
tjprescott: This has been fixed in the latest commit.

AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
sacheu: I have been thinking about that too. How to make the test method more concise. Compare to other languages, swift test cases are more verbose.  I think part of the problem is this pattern `            switch result {
            case .success:
                expectation.fulfill()
            case let .failure(error):
             XCTFail("some error string")
`

sacheu: Part of ^ being verbose because that makes debugger earlier. I can set breakpoints in success and failure cases and inspect the httpresponse when things failed.

AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
sacheu: I think `client.blah.putSimpleProduct(withGrouping: group)` make more sense.

AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
tjprescott: I think switching on the result is fine. My question is with the `client.omgwhatasuperlongvariablenamethisis.getResourceCollection`. Is this simply a result of the swagger using an operation group?  It would seem `client.getResourceCollection` would be preferrable.

sacheu: Yes. that is because of operation group. Per our discussion in our meeting earlier, we will separate operations even if there is only 1 operation group (as long as that operation group has a name).

tjprescott:
Updated #159 to indicate which tests pass with the PR and which still need to be worked on.

AutorestSwiftTest/AutoRestResourceFlatteningTest.swift
tjprescott: I agree, but I think I will separate this into a different issue, or let it be part of the PR for `test_resourceFlattening_putSimpleProductWithGroupingFlattenParameterGroup`

sacheu (approved):
Good to have more test cases.

PR #224, opened 2020-10-20
Add 4 test cases for Url.json PathItems

sacheu:
* Add 4 test cases for Url.json PathItems
* Improve AutoRest Swift to pull query param from Client property and optional Cilent Property
* Refactor: Remove adding "" in query param when required query params is empty


src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -158,11 +171,11 @@ struct KeyValueViewModel {
         - Parameter key: Key String in the Key value pair
         - Parameter value: the value string
      */
-    init(key: String, value: String) {
+    init(key: String, value: String, optional: Bool = false, implmentation: Implmentation = .options) {
```
tjprescott, line 174: Type here: implmentation -> implementation

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -65,10 +65,7 @@ struct OperationParameters {
             methodDecoding.append(param)
         }
 
-        // Add a blank key,value in order for Stencil generates an empty dictionary for QueryParams and PathParams constructor
-        if query.required.isEmpty {
-            query.required.append(KeyValueViewModel(key: "", value: "\"\""))
-        }
```
tjprescott, line 71: What's the rationale for removing this?

templates/OperationQueryParamFromClientPropertySnippet.stencil
```diff
@@ -0,0 +1,8 @@
+{% for param in op.params.query.optional %}
+{% if param.implementation == "client" %}
+if let {{ param.key }} = client.{{ param.key }} {
+    {% include "OperationOptionsDecodingSnippet.stencil" %}
+    queryParams.append("{{ param.key }}", {{ param.value}})
+}
+{% endif %}
```
tjprescott, line 7:
```suggestion
{% for param in op.params.query.optional %}
{% if param.implementation == "client" %}
if let {{ param.key }} = {{ param.path }} {
    {% include "OperationOptionsDecodingSnippet.stencil" %}
    queryParams.append("{{ param.key }}", {{ param.value}})
}
{% endif %}
```

Then you would not need to duplicate this code. Internally, you would set the path property as `self.path = "\(implementation.rawValue).{key}"`

templates/OperationUrlSnippet.stencil
```diff
@@ -5,7 +5,11 @@ guard let urlTemplate = "{{ op.request.path }}".removingPercentEncoding else {
 }
 let pathParams = [
     {% for param in op.params.path %}
+    {% if param.implementation == "client" %}
+        "{{ param.key }}" : client.{{ param.value}},
+    {% else %}
         "{{ param.key }}" : {{ param.value}},
+    {% endif %}
```
tjprescott, line 12: Same here, we can eliminate the if/else logic if we build up the value as a path.

test/integration/generated/url/Source/AutoRestUrlTestClient.swift
```diff
@@ -68,9 +68,9 @@ public final class AutoRestUrlTestClient: PipelineClient {
     // /// should contain value null
     public var globalStringQuery: String?
 
+    public lazy var queries: Queries = Queries(client: self)
     public lazy var paths: Paths = Paths(client: self)
     public lazy var pathitems: PathItems = PathItems(client: self)
```
tjprescott, line 73: I'm curious why the properties, upon being regenerated, change the ordering. There may be somewhere we are iterating through a collection that doesn't guarantee order. Not important for this PR, but it would be great to identify this and correct.

tjprescott (commented):
Some comments and questions.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Populate empty strings in query dictionary will result in the url query parameters. i.e. the url building logic does not skip when the key is "".  Moreover, putting "" in query array in generated code is not needed. The generated code will still compile.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -36,6 +36,11 @@ enum KeyValueDecodeStrategy: String {
     case dateTimeFromParam
 }
 
+enum Implmentation: String {
```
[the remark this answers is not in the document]
sacheu: Removed in new iteration.

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Remove in new iteration.

templates/OperationQueryParamFromClientPropertySnippet.stencil
sacheu: For some reason, {enum}.rawValue does not work in stencil. I guess rawValue is computed values. 
I have changed to adapt with your suggestion in new iteration.

templates/OperationUrlSnippet.stencil
sacheu: Removed in new iteration.

test/integration/generated/url/Source/AutoRestUrlTestClient.swift
sacheu: Add code to sort the shortcuts in Service Client Model.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -48,12 +53,13 @@ struct KeyValueViewModel {
...
     let needDecodingInMethod: Bool
-    // An enum indicates what kind of decoding strategy will be used in the method implementation
+    // An enum raw value indicates what kind of decoding strategy will be used in the method implementation
     let strategy: String
     // This is for Method Decoding stencil to pull in the value of the Constant when create a variable for the constant
     // Valid if the key-value is from a Constant schema. Otherwise, it will be nil
     let constantValue: String?
-
+    // An enum raw value indicatse whether the value is from client property or from options struct
+    let implementation: String
```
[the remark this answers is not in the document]
sacheu: In stencil templates, we need to tell if we generate code `client.` to pull values from client property. or `options.` to pull values from options struct.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -73,14 +74,20 @@ struct KeyValueViewModel {
         } else if let groupedBy = param.groupedBy?.name {
             self.init(key: name, value: "\(groupedBy).\(name)")
         } else if param.implementation == .client {
-            self.init(key: name, value: "client.\(name)")
+            self.init(
+                key: name,
+                value: param.required ? "client.\(name)" : name,
```
tjprescott, line 79: If the implementation is in the client, does it matter whether it is required or not? Wouldn't it still be client.property?

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -48,12 +48,13 @@ struct KeyValueViewModel {
...
     // Flag indicates if the key/value pair need decoding code in method to convert the variable into a String
     let needDecodingInMethod: Bool
-    // An enum indicates what kind of decoding strategy will be used in the method implementation
+    // An enum raw value indicates what kind of decoding strategy will be used in the method implementation
     let strategy: String
     // This is for Method Decoding stencil to pull in the value of the Constant when create a variable for the constant
     // Valid if the key-value is from a Constant schema. Otherwise, it will be nil
     let constantValue: String?
-
+    // A string to specify if the value comes from client property or from options struct or from signature
```
tjprescott, line 56: I would change this comment to say, "The full path to the value property."

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -123,6 +130,7 @@ struct KeyValueViewModel {
 
     init(signatureParameter: ParameterType, name: String) {
         self.key = name
+        self.path = signatureParameter.required ? name : "options.\(name)"
```
tjprescott, line 133: I'm not sure this logic is accurate. For example, if you flatten a body that has optional parameters, those optional parameters should appear in the signature. 

I think we should consult the `belongsInSignature` or `belongsInOptions` methods available on `ParameterType`.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -157,12 +165,14 @@ struct KeyValueViewModel {
...
+        - Parameter implementation: an enum indicates if the value is from Client property or from Options struct
      */
-    init(key: String, value: String) {
+    init(key: String, value: String, optional: Bool = false, path: String? = nil) {
         self.key = key
         self.value = value
-        self.optional = false
-
+        self.optional = optional
+        self.path = path ?? "options.\(key)"
```
tjprescott, line 175: The special case here makes me uneasy. If I have a key value, I feel that the path should point to the value, not the key. This might work since usually the key property and the  value property have the same name.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -48,8 +48,12 @@ struct OperationParameters {
...
             switch httpParam.in {
             case .query:
-                viewModel.optional ? query.optional.append(viewModel) : query.required
-                    .append(viewModel)
+                if param.implementation == .client {
+                    query.optionalClientProperty.append(viewModel)
+                } else {
+                    viewModel.optional ? query.optional.append(viewModel) : query.required
+                        .append(viewModel)
+                }
```
tjprescott, line 56: So right now it seems that `query` has two properties, `optionalClientProperty` and `optional`. Would it make more sense for `query.optional` to have a property that described where it comes from? `querty.optional.source` or something like that? I just worry that if there is another implementation location in the future, then we have ever-increasing `optionalFoo` properties.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -65,12 +69,9 @@ struct OperationParameters {
...
         }
 
-        // Add a blank key,value in order for Stencil generates an empty dictionary for QueryParams and PathParams constructor
-        if query.required.isEmpty {
-            query.required.append(KeyValueViewModel(key: "", value: "\"\""))
-        }
+        // Add a blank key,value in order for Stencil generates an empty dictionary for PathParams constructor
         if path.isEmpty {
-            path.append(KeyValueViewModel(key: "", value: "\"\""))
+            path.append(KeyValueViewModel(key: "", value: "\"\"", path: "\"\""))
```
tjprescott, line 74: This seems odd to me. I would expect this to generate something like `"".""` which is clearly not right.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -109,14 +110,17 @@ struct OperationParameters {
 struct Params {
     // Query Params/Header in initializer
     var required: [KeyValueViewModel]
-    // Query Params/Header need to add Nil check
+    // Query Params/Header which is pull from Options struct need to add Nil check
     var optional: [KeyValueViewModel]
+    // Query Params/Header which is pull from Optional Client property. It need to add Nil check
+    var optionalClientProperty: [KeyValueViewModel]
```
tjprescott, line 116: Not sure I agree with this approach. It seems the value side of the KeyValueViewModel should know whether the value is pulled from the client (`client.value`), the signature (`value`) or a constant (`"value"`)

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
```diff
@@ -69,6 +69,7 @@ struct ServiceClientFileViewModel {
         for key in namedOperationGroups.keys {
             namedOperationGroupShortcuts.append(KeyValueViewModel(key: key, value: key.lowercased()))
         }
+        namedOperationGroupShortcuts.sort(by: { $0.key > $1.key })
```
tjprescott, line 72: Will `namedOperationGroupShortcuts.sorted()` not suffice here?

templates/OperationOptionalQueryParamSnippet.stencil
```diff
@@ -0,0 +1,4 @@
+if let {{ param.key }} = {{ param.path }} {
+    {% include "OperationOptionsDecodingSnippet.stencil" %}
+    queryParams.append("{{ param.key }}", {{ param.value }})
```
tjprescott, line 3: I'm thinking that "value" might be losing its real meaning. As I look at this, I see that really, there should either be a constant or a property path (with or without a dot). Neither of those is captured by value.

```swift
{% if constant  $}
  queryParams.append("{{ param.key }}", "{{ param.constant }}")
{% else %}
  queryParams.append("{{ param.key }}", {{ param.path }})
{% endif %}
```

tjprescott (changes requested):
We may need to hop on a call to figure out the right way to expose the value side of the key-value model.

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Fixed.

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Removed.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Reverted.

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
sacheu: I need to implement KeyValueViewModel to follow 'Comparable' protocol if I switched to sorted.

templates/OperationOptionalQueryParamSnippet.stencil
sacheu: I change the KeyValue Model so that the param.value points to right value whether the value is from constant, option, signature parameter or client property. So we don't need the %if logic in stencil.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: I add a 'source' property in the KeyValue Model to indicate where the value of the KeyValueModel comes from.   And the 'source' property will be used in OperationOptionsSnippet and OperationSnippet.stencil so that we can unwrap optional options/client property before adding it to query parameter of the URL.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -48,11 +55,15 @@ struct KeyValueViewModel {
     let optional: Bool
     // Flag indicates if the key/value pair need decoding code in method to convert the variable into a String
     let needDecodingInMethod: Bool
-    // An enum indicates what kind of decoding strategy will be used in the method implementation
+    // An enum raw value indicates what kind of decoding strategy will be used in the method implementation
     let strategy: String
     // This is for Method Decoding stencil to pull in the value of the Constant when create a variable for the constant
     // Valid if the key-value is from a Constant schema. Otherwise, it will be nil
     let constantValue: String?
```
tjprescott, line 62: Are we actually using this? I thought the constant value was rolled into value?

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -114,9 +111,9 @@ struct Params {
     // Whether to 'var' or 'let' in generated code for the param declaration
     var declaration: String = "var"
 
-    init(from params: Params? = nil) {
-        self.required = params?.required ?? [KeyValueViewModel]()
-        self.optional = params?.optional ?? [KeyValueViewModel]()
+    init() {
+        self.required = [KeyValueViewModel]()
+        self.optional = [KeyValueViewModel]()
     }
```
tjprescott, line 117: Why was this removed?

templates/OperationOptionsSnippet.stencil
```diff
@@ -2,12 +2,10 @@
 if let options = options {
     // Query options
     {% for param in op.params.query.optional %}
-    if let {{ param.key }} = options.{{ param.key }} {
-        {% include "OperationOptionsDecodingSnippet.stencil" %}
-        queryParams.append("{{ param.key }}", {{ param.value}})
-    }
+    {% if param.source == "options" %}
+    {% include "OperationOptionalQueryParamSnippet.stencil" %}
+    {% endif %}
```
tjprescott, line 7: This feels wrong. We should not have to check that the source is options.

templates/OperationSnippet.stencil
```diff
@@ -15,6 +15,11 @@ public func {{ op.name }}(
     {% include "OperationMethodDecoding.stencil" %}
     {% include "OperationUrlSnippet.stencil" %}
     {% include "OperationQueryParamSnippet.stencil" %}
+    {% for param in op.params.query.optional %}
+    {% if param.source == "clientProperty" %}
+    {% include "OperationOptionalQueryParamSnippet.stencil" %}
+    {% endif %}
+    {% endfor %}
```
tjprescott, line 22: Same here. This feels like we are modeling this wrong if this is what our stencil has too look like.

tjprescott (changes requested):
It seems like things are getting more complex when they should be getting simpler. We should set up a meeting.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -73,14 +84,22 @@ struct KeyValueViewModel {
         } else if let groupedBy = param.groupedBy?.name {
             self.init(key: name, value: "\(groupedBy).\(name)")
         } else if param.implementation == .client {
-            self.init(key: name, value: "client.\(name)")
+            self.init(
+                key: name,
+                value: name,
+                optional: !param.required,
+                path: "client.",
+                source: .clientProperty
```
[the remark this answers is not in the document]
sacheu: yes.`path` string value is used in the generated code,  templates/OperationUrlSnippet.stencil

and source is used in templates/OperationSnippet.stencil and templates/OperationOptionsSnippet.stencil to add generated code to unwrap client property or option.

templates/OperationSnippet.stencil
sacheu: In my earlier iteration, I create 2 arrays to separate these 2 scenarios: 
* op.params.query.optional
* op.params.query.optionalClientProperty
But this was reverted

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: There is nowhere in the code which we pass in param when create this struct. So I remove the code to simplify the code.

templates/OperationUrlSnippet.stencil
```diff
@@ -5,7 +5,7 @@ guard let urlTemplate = "{{ op.request.path }}".removingPercentEncoding else {
 }
 let pathParams = [
     {% for param in op.params.path %}
-        "{{ param.key }}" : {{ param.value}},
+        "{{ param.key }}" : {{ param.path}}{{ param.value }},
```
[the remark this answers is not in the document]
sacheu: No. we need to add 'param.path' in the stencil because there is a scenario what we get the value from client property.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Reverted the change.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -48,11 +55,15 @@ struct KeyValueViewModel {
...
-    // An enum indicates what kind of decoding strategy will be used in the method implementation
+    // An enum raw value indicates what kind of decoding strategy will be used in the method implementation
     let strategy: String
     // This is for Method Decoding stencil to pull in the value of the Constant when create a variable for the constant
     // Valid if the key-value is from a Constant schema. Otherwise, it will be nil
     let constantValue: String?
+    // The full path to the value property
+    let path: String
+    // An enum raw value indicates the source of the valur property
+    let source: String
```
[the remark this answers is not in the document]
sacheu: Removed

test/integration/generated/url/Source/AutoRestUrlTestClient.swift
```diff
@@ -68,8 +68,8 @@ public final class AutoRestUrlTestClient: PipelineClient {
     // /// should contain value null
     public var globalStringQuery: String?
 
-    public lazy var paths: Paths = Paths(client: self)
     public lazy var pathitems: PathItems = PathItems(client: self)
+    public lazy var paths: Paths = Paths(client: self)
     public lazy var queries: Queries = Queries(client: self)
```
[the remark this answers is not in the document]
sacheu: It is now sorted.

src/AutorestSwift/View Models/KeyValueViewModel.swift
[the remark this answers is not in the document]
sacheu: Removed `source`

templates/OperationUrlSnippet.stencil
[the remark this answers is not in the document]
sacheu: Leave `path` property for now. Will refactor once AzureCore enhancement is available.

templates/OperationOptionsSnippet.stencil
sacheu: Removed

tjprescott (approved):
Great refactoring! This looks great. Thanks Sam

PR #225, opened 2020-10-20
Support encode unixTime in request body

sacheu:
* Support encode unixTime in request body
* Add 5 more test cases for body-integer.json

### DIFF
--- a/src/AutorestSwift/View Models/OperationViewModel.swift
+++ b/src/AutorestSwift/View Models/OperationViewModel.swift
@@ -103,6 +103,12 @@ struct OperationParameters {
     }
 }
 
+enum BodyParamStrategy: String {
+    case plain
+    case flattened
+    case unixTime
+}
+
 struct Params {
     // Query Params/Header in initializer
     var required: [KeyValueViewModel]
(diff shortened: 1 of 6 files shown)


src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -121,11 +121,6 @@ struct Params {
     }
 }
 
-enum BodyParamStrategy: String {
```
tjprescott, line 124: I think we should keep this since and just add unixTime as a case. While it is only used in the strategy property now, it could potentially be used elsewhere in the future and I wouldn't want to have magic strings all over the place (except in the stencil files where's there's no alternative)

tjprescott (changes requested):
Would like to keep the enum, but otherwise I think this change is fine.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Good Point. I remove that only because nowhere in the code was using this enum. I add the enum back and change the code to use it.

tjprescott (approved):

PR #226, opened 2020-10-20
Support $host in generated code

sacheu:
* Use $host to create base URL in init() of the Client generated code
* Create a convenience init() snippet
* Change all Test cases to use the convenience init() to create Client


scripts/testserver_codegen.sh
```diff
@@ -27,7 +27,7 @@ do
 	echo "autorest code generation succeed."
     else
 	echo "autorest code generation failed. Revert the generated code."
-	git restore ./test/integration/generated/$i
+	#git restore ./test/integration/generated/$i
```
tjprescott, line 30: Did you mean to comment this out?

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
```diff
@@ -62,10 +63,20 @@ struct ServiceClientFileViewModel {
...
+            } else {
+                globalParameters.append(ParameterViewModel(from: globalParameter))
+            }
         }
         self.globalParameters = globalParameters
+        if let hostValue = host {
+            self.host = "\"\(hostValue)\""
+        } else {
+            self.host = ""
+        }
```
tjprescott, line 79: I would suggest extracting this logic out to a private method since it's non-trivial. You can make it return a named tuple so you could have something like:

```swift
let globals = resolveGlobalParameters()
self.host = globals.host
self.globalParameters = globals.parameters 
```

test/integration/generated/body-file/Source/AutoRestSwaggerBatFileClient.swift
```diff
@@ -33,6 +33,25 @@ public final class AutoRestSwaggerBatFileClient: PipelineClient {
...
+    ///   - authPolicy: An `Authenticating` policy to use for authenticating client requests.
+    ///   - options: Options used to configure the client.
+    public convenience init(
+        authPolicy: Authenticating,
+        withOptions options: AutoRestSwaggerBatFileClientOptions
+    ) throws {
+        // let components = URLComponents()
+        // components.scheme = "http"
+        // components.host = self.host
+        // self.init(baseUrl: components.url, authPolicy: authPolicy, withOptions: options)
```
tjprescott, line 47: Why is this commented out? This was generated commented out? Looking at `ClientConvenienceInitSnippet` I don't see where this is coming from.

tjprescott (changes requested):
Some questions.

scripts/testserver_codegen.sh
sacheu: reverted

test/integration/generated/body-file/Source/AutoRestSwaggerBatFileClient.swift
sacheu: Regenerated the code and push to new iteration.

templates/ServiceClientFile.stencil
```diff
@@ -31,6 +31,9 @@ public final class {{ model.name }}: {{ model.protocols }} {
     public let options: {{ model.name }}Options
 
     // MARK: Initializers
+    {% if model.host != "" %}
+    {% include "ClientConvenienceInitSnippet.stencil" %}
+    {% endif %}
```
tjprescott, line 36: Could we not simply provide a default value for `baseUrl` in the regular init if the host parameter is supplied?

```swift
public init(
  baseUrl: URL = URL("blah"),
  authPolicy: Authenticating,
  withOptions options: Options
) throws {
```

tjprescott (commented):
One more question

templates/ServiceClientFile.stencil
sacheu: This does not work because URL("blah") will return an optional URL object. Therefore, we need to unwrap it because we can use it.

tjprescott: In that case, we can do this:

```swift
public init(
  baseUrl: URL? = nil,
  authPolicy: Authenticating,
  withOptions options: Options
) throws {
let hostValue = baseUrl ?? URL(String: hostValue)
guard hostValue != nil else {
  logger.error("Must supply URL or hostValue")
}
```

Something like that.

tjprescott: If $host is provided, is it provided per operation, or for the whole client? If for the whole client, should it not simply be a constant on the client?

sacheu: It is for the whole client. I am applying the whole constant in the convenience init(). I just create an URL object from the constant add a guard before calling the init(). This is an example of the generated code.

```
   public convenience init(
        authPolicy: Authenticating,
        withOptions options: AutoRestSwaggerBatFileClientOptions
    ) throws {
        guard let baseUrl = URL(string: "http://localhost:3000") else {
            fatalError("Unable to form base URL")
        }
        try self.init(
            baseUrl: baseUrl,
            authPolicy: authPolicy,
            withOptions: options
        )
    }
```

tjprescott:
```swift
public init(
        baseUrl: URL? = nil,
        authPolicy: Authenticating,
        withOptions options: AutoRestSwaggerBatFileClientOptions
    ) throws {
        let defaultHost = URL(string: "http://localhost:3000")
        let host = baseUrl ?? defaultHost
        guard host != nil else {
            fatalError("Unable to determine base URL. ")
        }
        try self.init(
            baseUrl: baseUrl,
            authPolicy: authPolicy,
            withOptions: options
        )
    }
```

If the swagger doesn't designate a $host property, then the initializer can be generated the same as it currently does. What I would like to avoid is another stencil file that essentially duplicates code from elsewhere.

sacheu: I was trying to avoid adding a if logic to swagger , that is why i create a new swagger for this scenario. 
Regardless, thanks for the feedback, I have upated new iteration per feedback.

tjprescott (approved):

PR #229, opened 2020-10-26
Add 14 test cases from Url.json

sacheu:
* Add 14 test cases from Url.json
* Add support for different delimiter in array
* Do not encode URL template from swagger, only the path items and query parameters
* Remove '!*'();:@&=+$,/?"' as allowable characters in query parameters
* Fix base64 encoding in path


src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -154,10 +160,9 @@ struct KeyValueViewModel: Comparable {
     /**
      Convert the type into String format in Swift
      */
-    private static func formatValueForType(type: AllSchemaTypes, value: String) -> String {
+    private static func formatValueForType(signatureParameter: ParameterType, value: String) -> String {
+        let type = signatureParameter.schema.type
```
tjprescott, line 164: A "swify"er signature might look like:

```suggestion
    private static func formatValue(forSignatureParameter signatureParameter: ParameterType, value: String) -> String {
        let type = signatureParameter.schema.type
```

templates/NamedOperationGroupFile.stencil
```diff
@@ -23,12 +23,11 @@ public final class {{ group.name }} {
         self.commonOptions = client.commonOptions
     }
 
-    public func url(forTemplate templateIn: String, withKwargs kwargs: [String: String]? = nil) -> URL? {
-        return client.url(forTemplate: templateIn, withKwargs: kwargs)
+    public func url(forTemplate templateIn: String, withKwargs kwargs: [String: String]? = nil, and addedParams: [QueryParameter]? = nil) -> URL? {
+        return client.url(forTemplate: templateIn, withKwargs: kwargs, and: addedParams)
```
tjprescott, line 27: What's the case for this?

templates/OperationRequestBodySnippet.stencil
```diff
@@ -22,8 +22,8 @@ guard let requestBody = try? JSONEncoder().encode(body) else {
     return
 }
 {% endif %}
-guard let requestUrl = url.appendingQueryParameters(queryParams) else {
-    self.options.logger.error("Failed to append query parameters to URL")
+guard let requestUrl = self.url(forTemplate: urlTemplate, withKwargs: pathParams, and: queryParams) else {
```
tjprescott, line 25: The `url` method in AzureCore does not have this `and` parameter. What is it for?

templates/Primitives+ExtensionFile.stencil
```diff
@@ -16,3 +16,12 @@ extension Int64: LocalizedError {
...
 
+extension Data {
+    func base64URLEncodedString() -> String {
+        let base64String = self.base64EncodedString()
+        let base64UrlString = base64String.replacingOccurrences(of: "/", with: "_")
+                        .replacingOccurrences(of: "+", with: "-")
+                        .replacingOccurrences(of: "=", with: "")
+        return base64UrlString
+    }
+}
```
tjprescott, line 27: Is this a workaround for the `+` issue? Should this be in AzureCore?

templates/ServiceClientFile.stencil
```diff
@@ -7,6 +7,11 @@
 // swiftlint:disable function_body_length
 // swiftlint:disable type_body_length
 
+extension CharacterSet {
+    static let urlQueryValueAllowed = urlQueryAllowed.subtracting(.init(charactersIn: "!*'();:@&=+$,/?"))
+    static let urlPathItemAllowed = urlPathAllowed.subtracting(.init(charactersIn: "!*'()@&=+$,/:"))
+}
+
```
tjprescott, line 14: Should this be in AzureCore?

templates/ServiceClientFile.stencil
```diff
@@ -79,6 +84,37 @@ public final class {{ model.name }}: {{ model.protocols }} {
...
+    private func appendingQueryParameters(url: URL, _ addedParams: [QueryParameter]) -> URL? {
+        guard !addedParams.isEmpty else { return url }
+        guard var urlComps = URLComponents(url: url, resolvingAgainstBaseURL: true) else { return nil }
+
+        let addedQueryItems = addedParams.map { name, value in URLQueryItem(name: name, value: value?.addingPercentEncoding(withAllowedCharacters: .urlQueryValueAllowed)) }
+        urlComps.percentEncodedQueryItems = addedQueryItems
+        return urlComps.url
+    }
+
```
tjprescott, line 117: This solution should be in AzureCore.

tjprescott (changes requested):
A lot of these changes should be made in AzureCore. I think it is probably fine to create an issue in AzureCore that lists the things that need to be ported and then merge this, since we should be able to extract them back out.

scripts/testserver_codegen.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+echo "== Compile AutoRest Swift =="
+
+make install > /dev/null 2>&1
+
+if [ $? -eq 0 ]; then
```
tjprescott, line 7: Bad rebase? This has been replaced by the python script, yes?

templates/Primitives+ExtensionFile.stencil
sacheu: Yes, this should be in AzureCore.

templates/ServiceClientFile.stencil
sacheu: Yes, this should be in AzureCore.

templates/OperationRequestBodySnippet.stencil
sacheu: I create a helper method so that it builts an URL based on urltemplate and query parameters. This is different from AzureCore , which has 1 method to build url and another to add query parameters.

templates/NamedOperationGroupFile.stencil
sacheu: I create a helper method so that it builts an URL based on urltemplate and query parameters. This is different from AzureCore , which has 1 method to build url and another to add query parameters.

tjprescott (approved):

PR #230, opened 2020-10-27
Add support for custom baseURl test cases

sacheu:
* Build request url from Request uri parameter in swagger file (when it is not $host). When it is $host, it should be using the PiplelineClient `baseurl`, which get initialized 
* Change the url method in ServiceClient to resolve parameterized values in the request uri
* Create a new stencil file to construct request url


src/AutorestSwift/Models/Complex/Parameter.swift
```diff
@@ -70,6 +70,36 @@ class Parameter: Value, CustomDebugStringConvertible {
         }
     }
 
+    var inPath: Bool {
+        if let x_in = extensions?["x-in"]?.value as? String,
+            x_in == ParameterLocation.path.rawValue {
+            return true
+        }
```
tjprescott, line 77: Swift variables don't use underscore.

src/AutorestSwift/Models/Complex/Parameter.swift
```diff
@@ -70,6 +70,36 @@ class Parameter: Value, CustomDebugStringConvertible {
...
+
+        guard let location = paramLocation else { return false }
+        return location == ParameterLocation.path
+    }
+
+    var inQuery: Bool {
+        if let x_in = extensions?["x-in"]?.value as? String,
+            x_in == ParameterLocation.query.rawValue {
+            return true
+        }
```
tjprescott, line 87: Is `x-in` just a duplicate way of specifying where a parameter should go? How does this differ from what we've been using?

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -43,19 +43,17 @@ struct OperationParameters {
...
+            } else if param.inHeader {
                 viewModel.optional ? header.optional.append(viewModel) : header.required
                     .append(viewModel)
-            case .path:
+            } else if param.inPath {
                 path.append(viewModel)
-            default:
+            } else {
                 continue
             }
```
tjprescott, line 58: Would like to understand the difference between `httpParam.in` and `x-in` extension.

src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -46,6 +47,10 @@ struct RequestViewModel {
     init(from request: Request, with operation: Operation) {
         // load HttpRequest properties
         let httpRequest = request.protocol.http as? HttpRequest
+        let uri = httpRequest?.uri ?? nil
+        // if the uri equals to '$host', the request url is retrieved from the PipelineClient baseUrl property.
+        // Do not need to save that in the RequestViewModel
+        self.uri = (uri != "{$host}") ? uri : nil
```
tjprescott, line 53: Aren't we using $host as the default baseUrl? How is this different?

templates/NamedOperationGroupFile.stencil
```diff
@@ -23,8 +23,8 @@ public final class {{ group.name }} {
         self.commonOptions = client.commonOptions
     }
 
-    public func url(forTemplate templateIn: String, withKwargs kwargs: [String: String]? = nil, and addedParams: [QueryParameter]? = nil) -> URL? {
-        return client.url(forTemplate: templateIn, withKwargs: kwargs, and: addedParams)
+    public func url(forTemplate templateIn: String, withHost hostIn: String? = nil, withKwargs kwargs: [String: String]? = nil, and addedParams: [QueryParameter]? = nil) -> URL? {
+        return client.url(forTemplate: templateIn, withHost: hostIn, withKwargs: kwargs, and: addedParams)
```
tjprescott, line 27: This also looks like something that should be in AzureCore.

tjprescott (approved):
Might need a call to understand how this is different from what we had before.

tjprescott (changes requested):
Wow, too early. I meant to mark this "Request Changes"

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: Removed x-in extension logic.

src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -47,10 +47,10 @@ struct RequestViewModel {
     init(from request: Request, with operation: Operation) {
         // load HttpRequest properties
         let httpRequest = request.protocol.http as? HttpRequest
-        let uri = httpRequest?.uri ?? nil
+        self.uri = httpRequest?.uri ?? nil
```
tjprescott, line 50: There's no reason to have `?? nil` since the point of `??` is to use the rhs when the lhs is nil

src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -47,10 +47,10 @@ struct RequestViewModel {
...
         // if the uri equals to '$host', the request url is retrieved from the PipelineClient baseUrl property.
         // Do not need to save that in the RequestViewModel
-        self.uri = (uri != "{$host}") ? uri : nil
+        // self.uri = (uri != "{$host}") ? uri : nil
```
tjprescott: Why is this commented out?

tjprescott (commented):
A couple additional comments.

src/AutorestSwift/Models/Complex/Parameter.swift
sacheu: Code is removed.

src/AutorestSwift/Models/Complex/Parameter.swift
sacheu: code is removed.

templates/NamedOperationGroupFile.stencil
sacheu: Open issue https://github.com/Azure/azure-sdk-for-ios/issues/534 for AzureCore

src/AutorestSwift/View Models/RequestViewModel.swift
sacheu: Have reverted the commented out code.

tjprescott (approved):

PR #237, opened 2020-10-29
Support code generation for body-string.json 

sacheu:
* If operation is a swift keyword (e.g. String, Enum), add 'Operation' as suffix 
* if name is a swift keyword, add 'Type' as suffix
* refactor package dependency in XCworkspace to not use xcodeproj. Use swift Package.swift instead.
* remove all the xcodeproj files in the repo


tjprescott (commented):
Looks like a lot of regenerated files. Have you updated the Wiki entry for creating new tests to account for no longer making XCodeProj?

sacheu:
Wiki is updated.

tjprescott (approved):

PR #239, opened 2020-10-29
Update for AzureSDK PR #515.

tjprescott:
Exposes context to user.

### DIFF
--- a/templates/OperationSendRequestSnippet.stencil
+++ b/templates/OperationSendRequestSnippet.stencil
@@ -15,6 +15,7 @@ let context = PipelineContext.of(keyValues: [
    {% endfor %}
 ])
 context.add(cancellationToken: options?.cancellationToken, applying: self.options)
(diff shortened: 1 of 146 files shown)


templates/OperationSendRequestSnippet.stencil
```diff
@@ -15,6 +15,7 @@ let context = PipelineContext.of(keyValues: [
    {% endfor %}
 ])
 context.add(cancellationToken: options?.cancellationToken, applying: self.options)
+context.merge(with: options?.context)
```
sacheu, line 18: The method `context.merge` sounds similar to `context.add`.  Especially if one calls context.add and next with context.merge.

tjprescott: Add only adds one thing. Merge combines all values from another context into the first one (in this case, any values the user has created)

PR #240, opened 2020-10-30
Add 10 test cases for body-string string operation

sacheu:
* Fix handing of byteArray in response body
* Fix missing signature parameters from request parameters
* Fix `path` when virtual params is from flatten group object
* Add 10 test cases for body-string string operation
* Remove body parameters from option struct
* Remove default value '=nil' in required and nullable parameter

### DIFF
--- a/src/AutorestSwift/Models/Complex/Request.swift
+++ b/src/AutorestSwift/Models/Complex/Request.swift
@@ -62,8 +62,20 @@ class Request: Metadata {
 extension Request {
     /// Retrieves a single body-encoded parameter, if there is one. Fails if there is more than one.
     var bodyParam: ParameterType? {
+        // Look for body parameters from signature parameters first. If none is found, look up in parameters.
+        var bodyParams = findBodyParams(parameters: signatureParameters)
+        if bodyParams.count == 0 {
+            bodyParams = findBodyParams(parameters: parameters)
+        }
+
+        // current logic only supports a single request per operation
+        assert(bodyParams.count <= 1, "Unexpectedly found more than 1 body parameters in request... \(name)")
+        return bodyParams.first
+    }
+
+    func findBodyParams(parameters: [ParameterType]?) -> [ParameterType] {
         var bodyParams = [ParameterType]()
-        for param in signatureParameters ?? [] {
+        for param in parameters ?? [] {
             if case let ParameterType.virtual(virtParam) = param {
                 let originalParam = ParameterType.regular(virtParam.originalParameter)
                 if !bodyParams.contains(originalParam) {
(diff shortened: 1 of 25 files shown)


src/AutorestSwift/Models/Complex/Request.swift
```diff
@@ -75,9 +87,7 @@ extension Request {
                 bodyParams.append(param)
             }
         }
-        // current logic only supports a single request per operation
-        assert(bodyParams.count <= 1, "Unexpectedly found more than 1 body parameters in request... \(name)")
```
tjprescott, line 79: This assertion is still relevant and shouldn't be removed.

templates/BodyParamSnippet.stencil
```diff
@@ -1,3 +1,3 @@
-{% if op.params.body.strategy == "plain" or op.params.body.strategy == "unixTime"  %}
-    {{ op.params.body.param.name }}: {{ op.params.body.param.type }}{{ op.params.body.param.defaultValue }},
+{% if op.params.body.strategy == "plain" or op.params.body.strategy == "unixTime" or op.params.body.strategy == "plainNullable" or op.params.body.strategy == "byteArray" %}
```
tjprescott, line 1: Is there a strategy for which this block doesn't apply? if so, it would probably be easier to change the logic to != for that.

test/integration/generated/body-string/Source/StringOperation.swift
```diff
@@ -334,11 +338,12 @@ public final class StringOperation {
 
     /// Set string value empty ''
     /// - Parameters:
-
+    ///    - empty :
```
tjprescott, line 341: Is there no text for the body param, or are we missing it because of the name switch?

tjprescott (changes requested):

src/AutorestSwift/Models/Complex/Request.swift
sacheu: The assert is not removed. It is just moved. See line 72 of the src/AutorestSwift/Models/Complex/Request.swift

```
	        // current logic only supports a single request per operation
        assert(bodyParams.count <= 1, "Unexpectedly found more than 1 body parameters in request... \(name)")
        return bodyParams.first
    }
```

templates/BodyParamSnippet.stencil
sacheu: Great point.

test/integration/generated/body-string/Source/StringOperation.swift
sacheu: It is missing due to name switch.

templates/BodyParamSnippet.stencil
sacheu: Try changing to `!=` but stencil can't resolve `{{ op.params.body.param.name }} ` and `{{ op.params.body.param.type }}` for some reason. So keep the `==` logic instead

test/integration/generated/body-string/Source/StringOperation.swift
sacheu: issue is fixed after upgrading autoerest testserver version.

tjprescott (approved):

PR #241, opened 2020-10-30
Azure core pipeline update (2)

tjprescott:
Fixes the issue with baseUrl being renamed to endpoint.


src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -73,7 +73,7 @@ struct KeyValueViewModel: Comparable {
             self.init(
                 key: name,
                 // if the parameter is $host, retrieve the value from client's 'baseUrl' property
```
sacheu, line 75: Update comment to `endpoint`.

sacheu (approved):

PR #245, opened 2020-11-06
skip adding endpoint and api-version to list of the global parameters

sacheu:
`endpoint` and `api-version` already have different logic in generated code. Do not add those to the 'global parameters' list.

### DIFF
--- a/src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
+++ b/src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
@@ -79,7 +79,8 @@ func resolveGlobalParameters(from model: CodeModel) -> ([ParameterViewModel], St
     for globalParameter in model.globalParameters ?? [] {
         if globalParameter.name == "$host" {
             host = globalParameter.value.clientDefaultValue
-        } else {
+            // `endpoint` and `api-version` have different logic in generated code. Do not add them to the list of global parameters
+        } else if !["endpoint", "api-version"].contains(globalParameter.serializedName) {
             globalParameters.append(ParameterViewModel(from: globalParameter))
         }
     }


src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
```diff
@@ -79,7 +79,8 @@ func resolveGlobalParameters(from model: CodeModel) -> ([ParameterViewModel], St
     for globalParameter in model.globalParameters ?? [] {
         if globalParameter.name == "$host" {
             host = globalParameter.value.clientDefaultValue
-        } else {
+            // `endpoint` and `api-version` have different logic in generated code. Do not add them to the list of global parameters
+        } else if !["endpoint", "api-version"].contains(globalParameter.serializedName) {
```
tjprescott, line 83: Should we add `apiVersion` to this list as well?

tjprescott (commented):
Question

src/AutorestSwift/View Models/ServiceClientFileViewModel.swift
sacheu: I did not because the logic we have for getting api version is looking for 'api-version' only. See this in CodeModel.swift. If we add `apiVersion`, we need to do in both places.

```
    func getApiVersion() -> String {
        if let constantSchema = globalParameter(for: "api-version")?.schema as? ConstantSchema {
            return constantSchema.value.value
        } else {
            return ""
        }
    }
```

sacheu: So is it expected for autorest swift to support both apiVersion and api-version in swagger?

tjprescott: Nope, this is good.

tjprescott (approved):

PR #246, opened 2020-11-09
Add test cases for enum operation in body-string.json

sacheu:
* if  a constant is `required` , encode the constant in request body.

### DIFF
--- a/src/AutorestSwift/View Models/ConstantViewModel.swift
+++ b/src/AutorestSwift/View Models/ConstantViewModel.swift
@@ -36,12 +36,14 @@ struct ConstantViewModel {
     // default value of the proeprty
     let defaultValue: ViewModelDefault
     let required: Bool
+    let serializedName: String
 
     /// Initialize from Value type (such as Property or Parameter)
     init(from schema: ConstantSchema, required: Bool) {
         let name = (schema.serializedName ?? schema.name).lowercasedFirst
         assert(!name.isEmpty)
         self.name = name
+        self.serializedName = (schema.serializedName ?? schema.name)
         self.comment = ViewModelComment(from: schema.description)
         self.defaultValue = ViewModelDefault(from: schema.value.value, isString: true, isOptional: false)
         self.required = required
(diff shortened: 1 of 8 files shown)


src/AutorestSwift/View Models/ConstantViewModel.swift
```diff
@@ -35,13 +35,17 @@ struct ConstantViewModel {
...
     /// Initialize from Value type (such as Property or Parameter)
-    init(from schema: ConstantSchema) {
+    init(from schema: ConstantSchema, required: Bool) {
         let name = (schema.serializedName ?? schema.name).lowercasedFirst
         assert(!name.isEmpty)
         self.name = name
+        self.serializedName = (schema.serializedName ?? schema.name)
         self.comment = ViewModelComment(from: schema.description)
         self.defaultValue = ViewModelDefault(from: schema.value.value, isString: true, isOptional: false)
+        self.required = required
```
tjprescott, line 49: What is the use case of an "optional constant"?  A nil constant is pointless, and by definition a constant can't change, so how does this even work?

templates/ModelDecoderEncoderSnippet.stencil
```diff
@@ -18,4 +23,9 @@ public func encode(to encoder: Encoder) throws {
     {% for property in model.properties %}
     {% include "ModelEncoderProperty.stencil" %}
     {% endfor %}
+    {% for constant in model.constants %}
+    {% if constant.required == true %}
+    try container.encode({{ constant.name }}, forKey: .{{ constant.name }})
+    {% endif %}
+    {% endfor %}
```
tjprescott, line 30: I guess I don't understand--if the swagger defines a constant, when would it ever not be sent over the wire? Looking at this, I would think all constants would be required and am not sure what role a non-required constant would play or why it would be in Swagger.

tjprescott (commented):
Some questions.

templates/ModelDecoderEncoderSnippet.stencil
sacheu: Yes. you are right, the check is not needed. I removed the 'required' check in sending Constants in http request. Stencil files and ConstantViewModel are also updated. Thanks @tjprescott

sacheu:
Optional Coverage:
Mobile Passed Test=0, Mobile Total Test=2, Coverage=0.00%
Pass Test=1, Total Test=41, Coverage=2.44%

Coverage:
Mobile Passed Test=73, Mobile Total Test=90, Coverage=81.11%
Pass Test=100, Total Test=598, Coverage=16.72%

tjprescott (approved):

PR #247, opened 2020-11-10
Add test case from body-bytes.json

sacheu:
* Add test case from body-bytes.json
* Do not encode data using JSONEncoder if the body is byteArrray


templates/BodyParamSnippet.stencil
```diff
@@ -1,3 +1,3 @@
-{% if op.params.body.strategy == "plain" or op.params.body.strategy == "unixTime" or op.params.body.strategy == "plainNullable" %}
+{% if op.params.body.strategy == "plain" or op.params.body.strategy == "unixTime" or op.params.body.strategy == "plainNullable" or op.params.body.strategy == "byteArray" %}
```
sacheu, line 1: I tried changing this to `{% if op.params.body.strategy != “flattened” %}` but stencil file does not work.

AutorestSwiftTest/AutoRestSwaggerBatByteTest.swift
```diff
@@ -0,0 +1,124 @@
...
+        wait(for: [expectation], timeout: 5.0)
+    }
+
+    func test_getInvalid() throws {
+        let expectation = XCTestExpectation(description: "Call byte.getInvalid")
+
+        client.byte.getInvalid { result, httpResponse in
+            switch result {
+            case let .success(data):
+                XCTAssertEqual(String(decoding: data, as: UTF8.self), "\"::::SWAGGER::::\"")
```
tjprescott, line 98: Why is the "get invalid" case expecting success?

templates/BodyParamSnippet.stencil
tjprescott: You could always do:
```swift
{% if op params.body.strategy == "flattened" %}{% else %}
...
```

tjprescott: Also, from here (https://stencil.fuller.li/en/latest/builtins.html#built-in-tags) it looks like this might work:
```swift
{% if not op.params.body.strategy == "flattened" %}
```

I'm not sure why the != operator isn't working, since the documentation says it should. I'd recommend posting an issue in the stencil repo, since that would be a bug.

sacheu: Thanks for the suggestion. I find the cause of the issue. 
We should check if op.params.body is not nil in templates/OperationSnippet.stencil before include templates/BodyParamSnippet.stencil. Otherwise, we will have op.params.body.strategy equals to ""

AutorestSwiftTest/AutoRestSwaggerBatByteTest.swift
sacheu: For the test case we get, `\"::::SWAGGER::::\"`  in http response body. 

In python test case, it expects a deserialize error. 

https://github.com/Azure/autorest.python/blob/autorestv3/test/vanilla/AcceptanceTests/test_byte.py

For swift, we are using Data object for byteArray, and we just since http response body is already in Data object, we just pass the Data back to the caller.

```
if [
                    200
                ].contains(statusCode) {
                    dispatchQueue.async {
                        completionHandler(.success(data), httpResponse)
                    }
                }
```

That is why we don't get any deserialize error. Plus the string ``\"::::SWAGGER::::\"` is a valid ASCII string, I am not sure why python test code expects a deserialize error.

tjprescott (approved):

sacheu:
Optional Coverage:
Mobile Passed Test=0, Mobile Total Test=2, Coverage=0.00%
Pass Test=1, Total Test=41, Coverage=2.44%

Coverage:
Mobile Passed Test=73, Mobile Total Test=90, Coverage=81.11%
Pass Test=106, Total Test=598, Coverage=17.73%

PR #248, opened 2020-11-12
Modifications to Generate Storage

tjprescott:
Modifications needed to support the creation of a Storage code model and generate the code.


src/AutorestSwift/Utils/String+Extensions.swift
```diff
@@ -89,13 +89,30 @@ extension String {
 
     var isReserved: Bool {
         let reserved = [
-            "Error",
-            "Int",
-            "String",
-            "Bool",
-            "Enum",
-            "Array"
+            "Error": true,
```
sacheu, line 92: What is the purpose of the flag ? And when it will be set to false?

test/integration/generated/body-integer/Source/IntOperation.swift
```diff
@@ -614,7 +614,7 @@ public final class IntOperation {
 
     /// Put max int32 value
     /// - Parameters:
-    ///    - max32 : int body
+    ///    - max32 :
```
sacheu, line 617: why the comment 'int bodoy' is removed?

test/integration/generated/url/Source/PathItems.swift
```diff
@@ -300,8 +300,8 @@ public final class PathItems {
         if let localStringQuery = options?.localStringQuery {
             queryParams.append("localStringQuery", localStringQuery)
         }
-        if let globalStringQuery = client.globalStringQuery {
-            queryParams.append("globalStringQuery", globalStringQuery)
+        if let GlobalStringQuery = client.GlobalStringQuery {
```
sacheu, line 303: i think one of the changes from 'serializedName' to 'name' break this. The name should be 'globalStringQuery' instead of captical 'GlobalStringQuery'

src/AutorestSwift/Utils/String+Extensions.swift
tjprescott: It would never be set to false. The whole purpose is to use efficient constant time hash lookups instead of O(n) lookups that you would have to do with arrays.

test/integration/generated/body-integer/Source/IntOperation.swift
tjprescott: This is something I'm looking into. I'm still trying to find where it came from in the Swagger.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -38,6 +38,11 @@ enum KeyValueDecodeStrategy: String {
     case unixTime
 }
 
+enum KeyValueKeyStrategy {
```
sacheu, line 41: This enum is no longer necessary

src/AutorestSwift/Models/Simple/Language.swift
```diff
@@ -46,16 +46,19 @@ class Language: Codable {
 
     struct PagingNames: Codable {
         /// name for the item property
-        var itemName: String
+        var itemName: String? = "value"
```
sacheu, line 49: Why we use 'value' as the default value for itemName?

sacheu: And if itemName has a default value, will it ever be 'nil'? Why we need to switch from `String` to `String?`?

src/AutorestSwift/View Models/ConstantViewModel.swift
```diff
@@ -35,7 +35,7 @@ struct ConstantViewModel {
     let comment: ViewModelComment
     // default value of the proeprty
     let defaultValue: ViewModelDefault
-    let serializedName: String
+    let serializedName: String?
```
sacheu, line 38: From the way how we set ` serializedName`, I think serializedName will not never nil.

src/AutorestSwift/View Models/KeyValueViewModel.swift
tjprescott: Yup! Removed.

src/AutorestSwift/View Models/ConstantViewModel.swift
sacheu: I think this change is not needed.

src/AutorestSwift/Models/Simple/Language.swift
tjprescott: https://github.com/Azure/autorest/blob/master/docs/extensions/readme.md#x-ms-pageable

The description calls out that the default is "value". Storage was not specifying this, so it was crashing. It has to be optional otherwise when we try to decode the YAML to Swift, it will crash because the "itemName" key isn't found.

src/AutorestSwift/View Models/ConstantViewModel.swift
tjprescott: Reverted.

sacheu (approved):

PR #249, opened 2020-11-12
Change to support body-number.json

sacheu:
* Serialize `Constant' in http request body
* Override JSONEncode to encode Decimal in string following https://stackoverflow.com/questions/62995804/swift-jsonencoder-number-rounding


src/AutorestSwift/Models/Complex/ParameterType.swift
```diff
@@ -113,6 +113,10 @@ enum ParameterType: Codable {
         return common.nullable ?? false
     }
 
+    var isConstantSchema: Bool {
+        return common.schema as? ConstantSchema != nil
+    }
+
```
tjprescott, line 119: It would seem there should be a more generic way to handle this. You should already be able to say

```swift
if property.schema is ConstantSchema {
...
}
```

Otherwise I feel this should be something where you pass in the type you want to compare it to so we don't have a bunch of `isConstantSchema`, `isDateSchema` etc.

templates/Primitives+ExtensionFile.stencil
```diff
@@ -25,3 +25,32 @@ extension Data {
...
+}
+
+extension KeyedDecodingContainer {
+    func decode(_ type: Decimal.Type, forKey key: K) throws -> Decimal {
+        try decode(String.self, forKey: key).decimal(codingPath, key: key)
+    }
+    func decodeIfPresent(_ type: Decimal.Type, forKey key: K) throws -> Decimal? {
+        try decodeIfPresent(String.self, forKey: key)?.decimal(codingPath, key: key)
+    }
+}
```
tjprescott, line 56: This all needs to go into AzureCore. Please open an issue to port Primitives+ExtensionFile.

tjprescott (changes requested):

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -78,7 +78,7 @@ struct KeyValueViewModel: Comparable {
                 path: "client."
             )
         } else {
-            self.init(key: name, value: "")
+            self.init(key: name, value: "?????")
```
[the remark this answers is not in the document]
sacheu: This change is reverted in new iteration.

src/AutorestSwift/Models/Complex/ParameterType.swift
sacheu: Good point.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -80,8 +84,12 @@ struct OperationParameters {
...
         // Set the body param, if applicable
         if let bodyParam = operation.request?.bodyParam {
-            let bodyParamName = operation.request?.bodyParamName(for: operation)
-
+            var bodyParamName: String?
+            if body.count > 0 {
+                bodyParamName = body.first?.value
+            } else {
+                bodyParamName = operation.request?.bodyParamName(for: operation)
+            }
```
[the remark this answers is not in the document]
sacheu: Added assertion

templates/Primitives+ExtensionFile.stencil
sacheu: Code is removed.

tjprescott (approved):

PR #250, opened 2020-11-12
Change to support header.json

sacheu:
* Use name instead of serialized name if implementation is method
* Use signatureParameter name


tjprescott (approved):

src/AutorestSwift/Models/Complex/Schema.swift
```diff
@@ -124,6 +124,8 @@ class Schema: Codable, LanguageShortcut {
                 fatalError("Type mismatch. Expected constant type but got \(self)")
             }
             swiftType = constant.valueType.swiftType()
+        case AllSchemaTypes.duration:
+            swiftType = "TimeInterval"
```
tjprescott, line 128: "TimeInterval" is, by definition, always in seconds. Is "duration" intended to always be seconds?

templates/Primitives+ExtensionFile.stencil
```diff
@@ -25,3 +25,32 @@ extension Data {
...
+}
+
+extension KeyedDecodingContainer {
+    func decode(_ type: Decimal.Type, forKey key: K) throws -> Decimal {
+        try decode(String.self, forKey: key).decimal(codingPath, key: key)
+    }
+    func decodeIfPresent(_ type: Decimal.Type, forKey key: K) throws -> Decimal? {
+        try decodeIfPresent(String.self, forKey: key)?.decimal(codingPath, key: key)
+    }
+}
```
tjprescott, line 56: Needs an AzureCore issue filed to up-level these extensions

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -185,26 +155,84 @@ struct KeyValueViewModel: Comparable {
...
+        case .number:
+            if let numberSchema = parameter.schema as? NumberSchema {
+                return (numberSchema.swiftType() == "Decimal") ? .decimal : .number
+            } else {
+                return .number
+            }
+        default:
+            return .default
+        }
+    }
```
tjprescott, line 234: Is there a reason this can't just be a computed property directly on `ParameterType`?

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -185,26 +155,84 @@ struct KeyValueViewModel: Comparable {
...
+                if swiftType == "Decimal" {
+                    return "\(swiftType)(\(constantValue))"
+                } else {
+                    return "String(\(swiftType)(\(constantValue)))"
+                }
+            default:
+                return "String(\(constantValue))"
+            }
+        }
+    }
```
tjprescott, line 206: Is there a reason this isn't just a function directly on `ConstantSchema`?

tjprescott (changes requested):
A couple questions.

templates/Primitives+ExtensionFile.stencil
sacheu: This should be removed too. Will removed them.

src/AutorestSwift/Models/Complex/Schema.swift
sacheu: From the example, i see, yes TimeInterval is in seconds

tjprescott: I mean, is "duration" as spelled out in ModelerFour, always guaranteed to be in seconds?

src/AutorestSwift/View Models/ParameterViewModel.swift
```diff
@@ -41,7 +41,7 @@ struct ParameterViewModel {
         if let name = specificName, !name.isEmpty {
             self.name = name
         } else {
-            self.name = param.serializedName ?? param.name
+            self.name = /* param.serializedName ?? */ param.name
```
tjprescott: Why is this commented out? Should it be removed?

templates/Primitives+ExtensionFile.stencil
tjprescott: Even the extension for Decimal?

sacheu: Yes. I have removed them. I add code to convert a Decimal into a String (instead of doing it in KeyedEncodingContainer/KeyedDecodingContainer)

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Good suggestion. Have moved the code to `ParameterType`

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Good suggestion. Have moved the code to ConstantSchema

src/AutorestSwift/View Models/ParameterViewModel.swift
sacheu: This change is outdated. And I check the latest iteration, it has been removed.

test/integration/generated/url/Source/Queries.swift
```diff
@@ -1169,7 +1169,7 @@ public final class Queries {
 
         // Construct headers
         var headers = HTTPHeaders()
-        headers["Accept"] = "application/json"
+        headers["accept"] = "application/json"
```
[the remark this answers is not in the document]
sacheu: This change is outdated. And I check the latest iteration, the regenerated code is updated.

src/AutorestSwift/Models/Complex/Schema.swift
sacheu: The test server expects something like '**P123DT22H14M12.011S**'

which reads as 123 days 22 hr 14 min 12 sec.


```
      case "duration":
        if (scenario === "valid") {
          //For some reason moment.js doesn't quite get the right time value out (due to what looks like floating point issues)
          //so we have to check for two possible times
          if (value === "**P123DT22H14M12.011S**" || value == "P123DT22H14M12.010999999998603S") {
            coverage['HeaderParameterDurationValid']++;
            res.status(200).end();
            break;
          }
        }
```

sacheu: Change to use 'DateInterval' for 'duration'

src/AutorestSwift/Models/Complex/ConstantSchema.swift
```diff
@@ -112,4 +112,36 @@ class ConstantSchema: Schema {
 
         try super.encode(to: encoder)
     }
+
+    /**
+     Convert the Constant into String format in Swift
+     */
+    public func formatValue(isSkipUrlEncoding: Bool) -> String {
```
tjprescott, line 119: For documenting public functions, you should right-click, "Show Code Actions" > "Add Documentation". This will pre-populate the docstring with the fields in the right format.

tjprescott: I would rename `isSkipUrlEncoding` to just `skipUrlEncoding`.

src/AutorestSwift/Models/Complex/ParameterType.swift
```diff
@@ -151,6 +151,62 @@ enum ParameterType: Codable {
     internal func belongsInOptions() -> Bool {
         return common.belongsInOptions()
     }
+
+    /**
+     Convert the type into String format in Swift
+     */
+    public func formatValue(value: String) -> String {
```
tjprescott, line 158: Same comment on docstring, but this is an instance where you would use either of these signatures:

```swift
public func format(value: String) -> String {...}
public func formatValue(_ value: String) -> String {...}
```

For consistency with the other instance of this method, I would recommend the second option.

tjprescott (commented):
Very close. A couple small comments.

tjprescott (approved):

PR #252, opened 2020-11-13
Refactor param.serializedName

sacheu:
* Revist places in the code which uses param.serializedName if possible.
* For cases which use param.name, add lowercaseFirst to make sure first character is lower case per Swift coding standing
* For cases which use param.serializedName, change to use "" as default value so that we can validate our assume where are the cases we should always use serializedName


src/AutorestSwift/View Models/ConstantViewModel.swift
```diff
@@ -39,10 +39,10 @@ struct ConstantViewModel {
 
     /// Initialize from Value type (such as Property or Parameter)
     init(from schema: ConstantSchema) {
-        let name = (schema.serializedName ?? schema.name).lowercasedFirst
+        let name = schema.name.lowercasedFirst
         assert(!name.isEmpty)
         self.name = name
-        self.serializedName = (schema.serializedName ?? schema.name)
+        self.serializedName = schema.serializedName ?? schema.name
```
tjprescott, line 45: @daviwil is this a good assumption? If serialized name is omitted, I suppose it would make sense to use the regular name to send it over the wire.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -61,15 +61,14 @@ struct KeyValueViewModel: Comparable {
...
-            self.init(param: param, constantSchema: constantSchema, name: name)
-        } else if let signatureParameter = operation.signatureParameter(for: name) {
+            self.init(param: param, constantSchema: constantSchema)
+        } else if let signatureParameter = operation.signatureParameter(for: param.name) {
             self.init(signatureParameter: signatureParameter)
         } else if let groupedBy = param.groupedBy?.name {
-            self.init(key: name, value: "\(groupedBy).\(name)")
+            self.init(key: param.name, value: "\(groupedBy).\(param.name)")
         } else if param.implementation == .client {
+            let name = param.name.lowercasedFirst
```
tjprescott, line 71: It seems a little random where we are applying `lowercasedFirst`. Can we make it so that the getter for name is always lowercasedFirst?

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -78,14 +77,15 @@ struct KeyValueViewModel: Comparable {
                 path: "client."
             )
         } else {
-            self.init(key: name, value: "?????")
+            self.init(key: param.name, value: "")
```
tjprescott, line 80: Is this a situation we expect? If not, perhaps an assert would be wise here.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -78,14 +77,15 @@ struct KeyValueViewModel: Comparable {
...
         }
     }
 
-    private init(param: ParameterType, constantSchema: ConstantSchema, name: String) {
+    private init(param: ParameterType, constantSchema: ConstantSchema) {
         self.optional = false
         self.needDecodingInMethod = false
         self.path = ""
+        let name = (param.paramLocation == .header) ? (param.serializedName ?? "") : param.name
```
tjprescott, line 88: I'm wondering how much we can really special-case this. In any case, an assert would be appropriate here.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -117,7 +117,7 @@ struct KeyValueViewModel: Comparable {
     }
 
     private init(signatureParameter: ParameterType) {
-        self.key = signatureParameter.serializedName ?? signatureParameter.name
+        self.key = signatureParameter.serializedName ?? ""
```
tjprescott, line 120: Assert for an empty string key.

tjprescott (changes requested):
I think we need some asserts to check assumptions (so we don't just end up with code that won't compile) and a more consistent way to apply `lowercasedFirst`. Otherwise, this looks fine to me.

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: I have tried that approach (that the getter for name is always lowercasedFirst) first, but it breaks many other cases where we need to name to be upper case as First letter.  E.g. Model name, Options name, Enum Names, etc.
So that is why call `lowercasedFirst` at the places we needed.

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: The key in this case is the Http header value like Content-Type, Accept (all are upper cases.

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Added an assert.

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Added an assert.

src/AutorestSwift/View Models/KeyValueViewModel.swift
tjprescott: The assert would be for a blank key.

src/AutorestSwift/Models/Complex/ParameterType.swift
```diff
@@ -57,6 +57,10 @@ enum ParameterType: Codable {
         return common.name
     }
 
+    var swiftVariableName: String {
+        return common.swiftVariableName
+    }
```
tjprescott, line 62: Would it make sense to have `swiftModelName` as well? That would make the logic a bit more declarative and ensure the names are cased appropriately. We might also put an assert in here that ensures that `none-of-these` make it through. 

I would also be okay with dropping the "swift" prefix and just calling them `variableName` and `modelName`.

src/AutorestSwift/View Models/KeyValueViewModel.swift
```diff
@@ -101,11 +102,16 @@ struct KeyValueViewModel: Comparable {
...
 
-    private init(param: ParameterType, constantSchema: ConstantSchema, name: String) {
+    private init(param: ParameterType, constantSchema: ConstantSchema) {
         self.optional = false
         self.needDecodingInMethod = false
         self.path = ""
-        self.key = name
+        if param.paramLocation == .header {
+            assert(!(param.serializedName?.isEmpty ?? true))
+            self.key = param.serializedName ?? ""
```
tjprescott, line 111: This seems simpler.
```suggestion
            self.key = param.serializedName ?? ""
            assert(self.key != "", "Unexpected empty key.")
```

tjprescott (changes requested):
Like the declarative variable name helper. A couple comments.

src/AutorestSwift/Models/Complex/ParameterType.swift
sacheu: Switch to use modelName where appropriate.

src/AutorestSwift/View Models/KeyValueViewModel.swift
sacheu: Added an assert

tjprescott (approved):

PR #254, opened 2020-11-16
Fix to enable required-optional.json & body-array.json

sacheu:
Fix Autorest Swift to generate code for required-optional.json & body-array.json


tjprescott (approved):
LGTM

PR #261, opened 2020-11-19
Support url multi collection format

sacheu:
* Support explode property in httpParameter for array.

For each item in the array, if it is not nill, it should generate a query parameter (eg. arrayQuery) and the element value.

Example:

[](http://localhost:3000/queries/array/multi/string/valid?arrayQuery=ArrayQuery1&arrayQuery=begin%21%2A%27%28%29%3B%3A%40%20%26%3D%2B%24%2C%2F%3F%23%5B%5Dend&arrayQuery=&arrayQuery=)

* Add 3 test cases for url-multi-collectionFormat.json

Optional Coverage:
Pass Test=8, Total Test=42, Coverage=19.05%

Coverage:
Mobile Passed Test=86, Mobile Total Test=87, Coverage=98.85%
Pass Test=153, Total Test=598, Coverage=25.59%


src/AutorestSwift/Models/Complex/ParameterType.swift
```diff
@@ -176,7 +184,11 @@ enum ParameterType: Codable {
              .sealedChoice:
             return "\(value).rawValue"
         case .array:
-            return "\(value).map { String($0) }.joined(separator: \"\(delimiter)\") "
+            if explode {
+                return "\(value)"
+            } else {
+                return "\(value).map { String($0) }.joined(separator: \"\(delimiter)\") "
+            }
```
tjprescott, line 191: I'm confused here. If I had `arrayQuery=["A", "B", "C"]` I would expect the following:

```swift
if explode:
  return "arrayQuery=A&arrayQuery=B&arrayQuery=C"
else
  return "A,B,C"
```

tjprescott (commented):
One question. Otherwise LGTM.

src/AutorestSwift/Models/Complex/ParameterType.swift
sacheu: Your expectation is correct. The construction of the query parameters is done in the stencil file:

from templates/OperationOptionsSnippet.stencil

```

{% for param in op.params.explodeQuery.optional %}
if let {{ param.key }} = {{ param.path }}{{ param.key }} {
    {{ param.value}}.forEach {
        queryParams.append("{{ param.key }}", $0)
    }
}
{% endfor %}

```

tjprescott (approved):

PR #264, opened 2020-11-20
Autorest Swift changess to Support body-date.json, body-datetime.json, body-datetimerfc1123.json

sacheu:
* Encode Date,DateTime, and DateTime to string
* Add stencil file to decode when response is  Date,DateTime, and DateTime
* Refactor in KeyValueViewModel


src/AutorestSwift/View Models/ResponseViewModel.swift
```diff
@@ -42,13 +42,30 @@ enum ResponseBodyType: String {
     /// Service returns a byteArray body
     case byteArrayBody
 
-    static func strategy(for input: String, and type: AllSchemaTypes) -> ResponseBodyType {
+    case dateBody
+    case dateTimeBody
+    case datetimeRfc1123Body
+
+    static func strategy(for input: String, and type: AllSchemaTypes, schema: Schema) -> ResponseBodyType {
```
tjprescott, line 49: If we now need more information than just the type enum, we should just make this:
```swift
static func strategy(for input: String, and schema: Schema) -> ResponseBodyType
```

Because type is a property of Schema.

tjprescott (changes requested):
One comment. Otherwise looks good.

tjprescott (approved):

Issue #192, opened 2020-09-15
Conditionally create code for apiversion


sacheu:
The latest generated AzureCommunicationChat client did not generate the API version field correctly:

```swift
     /// API version of the  to invoke. Defaults to the latest.
     public enum ApiVersion: String {
        /// API version ""
        case v = ""
 
         /// The most recent API version of the
         public static var latest: ApiVersion {
            return .v
         }
     }

```

Should be:
```swift
     /// API version of the  to invoke. Defaults to the latest.
     public enum ApiVersion: String {
        /// API version "2020-09-21-preview2"
        case v20200921preview2 = "2020-09-21-preview2"
 
         /// The most recent API version of the
         public static var latest: ApiVersion {
            return .v20200921preview2
         }
     }
```

We need to:

- [ ] Fix the generation of ApiVersion enum when specified in Swagger
- [ ] Do not generate the ApiVersion enum if apiVersion is not specified in Swagger
- [ ] Write a simple Swagger to test both cases and add that to our test suite. It might make sense to put this test Swagger in the testserver.

tjprescott:
Since apiVersion is required by the ClientOptions protocol, we will leave this as-is.

PR #270, opened 2020-11-23
Fix AutorestSwift to generate Chat SDK

sacheu:
* Use 'api version' from operations
* Fix 'endpoint' parameter to use url absoluteString
* Use 'variable name' to look up signature parameter
* Use 'self.client' in PageCollection.


test/integration/generated/body-array/Source/AutoRestSwaggerBatArrayClient.swift
```diff
@@ -24,12 +24,12 @@ extension CharacterSet {
...
-        /// API version ""
-        case v = ""
+        /// API version "1.0.0"
+        case v100 = "1.0.0"
 
         /// The most recent API version of the
         public static var latest: ApiVersion {
-            return .v
+            return .v100
         }
```
tjprescott, line 33: Perhaps we should replace the `.` with `_`. Otherwise 1.0.0` would translate to the same thing as `10.0`, which would be problematic.

src/AutorestSwift/Models/Complex/CodeModel.swift
```diff
@@ -60,6 +60,14 @@ class CodeModel: Codable, LanguageShortcut {
...
             return constantSchema.value.value
         } else {
+            // If there is not `api-version` global parameter is defined, look for apiVersions in Operation for api version
+            for operationGroup in operationGroups {
+                for operation in operationGroup.operations {
+                    if let apiVersions = operation.apiVersions {
+                        return apiVersions[0].version
+                    }
+                }
+            }
```
tjprescott, line 70: This collection defines, I believe, the apiVersion(s) that an operation supports, which is not the same as getting the api version for the Swagger. Where does the Chat swagger define its api version?

tjprescott (changes requested):
Couple of issues.

test/integration/generated/body-array/Source/AutoRestSwaggerBatArrayClient.swift
sacheu: Fixed.

src/AutorestSwift/Models/Complex/CodeModel.swift
sacheu: Revert the change here.

test/integration/generated/body-array/Source/Generated/AutoRestSwaggerBatArrayClient.swift
```diff
@@ -24,12 +24,12 @@ extension CharacterSet {
...
     public enum ApiVersion: String {
-        /// API version "1.0.0"
-        case v100 = "1.0.0"
+        /// API version ""
+        case v = ""
 
         /// The most recent API version of the
         public static var latest: ApiVersion {
-            return .v100
+            return .v
```
tjprescott: So, does this not have an api version for the Swagger itsefl? If so, we should omit this entirely as described: https://github.com/Azure/autorest.swift/issues/192

tjprescott (commented):
One final comment. Would also be fine addressing https://github.com/Azure/autorest.swift/issues/192 in a separate PR, though it seems easy enough to do here.

test/integration/generated/body-array/Source/Generated/AutoRestSwaggerBatArrayClient.swift
sacheu: Yes. the testserver swaggers do not have apiVersion defined in global parameters. They only have apiversion defined in each operation. That was why I had the logic to look for apiVersion in operation in earlier iteration. But I have removed that in latest iteration.

Regarding #192, I did try to remove blank api version in generated code when first working on this PR.
But the code won't compile because of ClientOptions in AzureCore expects apiVersion in the protocol.

https://github.com/Azure/azure-sdk-for-ios/blob/50fbd64b19d50a40ecd045ee623cb48e1939e45f/sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift

image

tjprescott (approved):

PR #273, opened 2020-11-24
Fix Patch operation in Chat sdk

sacheu:
Use name instead of serialize name in search 'ContentType' parameter in request parameter when generated Patch operation

### DIFF
--- a/src/AutorestSwift/View Models/RequestViewModel.swift
+++ b/src/AutorestSwift/View Models/RequestViewModel.swift
@@ -60,7 +60,7 @@ struct RequestViewModel {
 
         // Determine which kind of request body snippet to render
         if method == "patch" {
-            if let contentTypeParameter = request.parameter(for: "Content-Type"),
+            if let contentTypeParameter = request.parameter(for: "contentType"),
                 let constantSchema = contentTypeParameter.schema as? ConstantSchema,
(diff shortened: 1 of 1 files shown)


src/AutorestSwift/View Models/RequestViewModel.swift
```diff
@@ -60,7 +60,7 @@ struct RequestViewModel {
 
         // Determine which kind of request body snippet to render
         if method == "patch" {
-            if let contentTypeParameter = request.parameter(for: "Content-Type"),
+            if let contentTypeParameter = request.parameter(for: "contentType"),
```
tjprescott, line 63: So the definition of this header in AzureCore is `Content-Type`. This suggests to me that the chat Swagger is wrong and not autorest.

tjprescott (changes requested):

src/AutorestSwift/View Models/RequestViewModel.swift
sacheu: In code model, the name is `contentType` and the serialized name is `Content-Type`. But in the request.parameter() method, we look for `parameter name` instead of `parameter serialized name` in the search.  The change to search in 'parameter name '  is introduced by a recent change. That is why it is broken.

tjprescott (approved):

PR #277, opened 2020-11-25
Fix Autorest Swift to generate compile code for paging.json

sacheu:
* Fix an issue in Model4 which create an ObjectSchema with `usage` and `serializationFormats` missing.
* In paging.json , it has 
```
          {
            "name": "maxresults",
            "in": "header",
            "required": false,
            "type": "integer",
            "format": "int32",
            "description": "Sets the maximum number of items to return in the response.",
            "x-ms-parameter-grouping": {
                "postfix": "Options"
            }
          },
```
and this translates to code model, 
```
          schema: *ref_2
          originalParameter:
            - !<!Parameter> &ref_33
              schema: *ref_2
              groupedBy: !<!Parameter> &ref_5
                schema: *ref_3
                implementation: Method
                language: !<!Languages> 
                  default:
                    name: pagingGetMultiplePagesOptions
                    description: Parameter group
                protocol: !<!Protocols> {}
              implementation: Method
              language: !<!Languages> 
                default:
                  name: maxresults
                  description: Sets the maximum number of items to return in the response.
                  serializedName: maxresults
              protocol: !<!Protocols> 
                http: !<!HttpParameter> 
                  in: header
          serializedName: maxresults
          language: !<!Languages> 
            default:
              name: maxresults
              description: Sets the maximum number of items to return in the response.
          protocol: !<!Protocols> {}
```

Since we already have an Option for each method, this `pagingGetMultiplePagesOptions` is duplicate of the `Option` struct that AutorestSwift created.   So we will put all the optional properties from `pagingGetMultiplePagesOptions` to the `Option` struct that AutorestSwift created.

* But for the property in `pagingGetMultiplePagesOptions` which is in `path`, the property should be **required**.  We need to add that property in the signature of the method instead of generating that in the Option struct.

See https://github.com/Azure/autorest.swift/pull/278 for generated code. All existing code does not change due to this PR.


src/AutorestSwift/View Models/ClientMethodOptionsViewModel.swift
```diff
@@ -53,8 +53,25 @@ struct ClientMethodOptionsViewModel {
...
+            for property in group.properties ?? [] {
+                switch property {
+                case let .grouped(groupProperty):
+                    for param in groupProperty.originalParameter where param.paramLocation != .path {
+                        properties.append(PropertyViewModel(from: property.value))
+                    }
+                default:
+                    continue
+                }
+            }
```
tjprescott, line 74: Is there an example of this? I think I get it, but the special casing concerns me somewhat.

src/AutorestSwift/View Models/ObjectViewModel.swift
```diff
@@ -96,7 +96,7 @@ struct ObjectViewModel {
             // the source of flattened parameters should not be included in the view model
 
             // FIXME: This assumption no longer holds for storage and should be revisited
-            assert(property.originalParameter.count <= 1, "Expected, at most, one original parameter.")
+            // assert(property.originalParameter.count <= 1, "Expected, at most, one original parameter.")
```
tjprescott, line 99: Does this assert fail for paging?

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -109,8 +109,25 @@ struct OperationParameters {
...
+                            signatureParameterViewModel
+                                .append(ParameterViewModel(from: param))
+                        }
+                    default:
+                        continue
+                    }
+                }
+            } else if param.belongsInSignature() {
+                signatureParameterViewModel.append(ParameterViewModel(from: param))
+            }
```
tjprescott, line 130: I wonder if this check should be done when the model is created. It seems strange to need to pass the model in here. If these params that should be in the options model were processed and popped earlier, we wouldn't need the reference to the model here.

tjprescott (commented):
A few questions, but ultimately, if it works, it works.

src/AutorestSwift/Code Generator/UsageSchema.swift
```diff
@@ -28,5 +28,5 @@ import Foundation
 
 /// Protocol that delinates models with a usage property
 protocol UsageSchema {
-    var usage: [SchemaContext] { get }
+    var usage: [SchemaContext]? { get }
```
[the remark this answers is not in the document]
sacheu: This is an issue I found in Autorest. It sends us an ObjectSchema which does not have `usage` and `serializationFormats`

https://github.com/Azure/autorest/issues/3708

Above is changed to work around the Model4 issue. This issue was hard to debug as AutorestSwift just failed to generate any code. It took some time to find out it is because of missing  `usage` and `serializationFormats`. I have filed an issue to AutoRest

src/AutorestSwift/View Models/ObjectViewModel.swift
sacheu: Yes. I see when generating code for paging.json

```
WARNING (PreCheck/CheckDuplicateSchemas): Checking for duplicate schemas, this could take a (long) while.  Run with --verbose for more detail.
Assertion failed: Expected, at most, one original parameter.: file AutorestSwift/ObjectViewModel.swift, line 175
Process() cancelled due to failure
```

src/AutorestSwift/View Models/ClientMethodOptionsViewModel.swift
sacheu: test swagger paging.json in an example.

In the swagger, it specifies 
```
            "name": "maxresults",
            "in": "header",
            "required": false,
            "type": "integer",
            "format": "int32",
            "description": "Sets the maximum number of items to return in the response.",
            "x-ms-parameter-grouping": {
                "postfix": "Options"
            }
          },
```

And this translates to code model:
```
         originalParameter:
            - !<!Parameter> &ref_33
              schema: *ref_2
              groupedBy: !<!Parameter> &ref_5
                schema: *ref_3
                implementation: Method
                language: !<!Languages> 
                  default:
                    name: pagingGetMultiplePagesOptions
                    description: Parameter group
                protocol: !<!Protocols> {}
              implementation: Method
              language: !<!Languages> 
                default:
                  name: maxresults
                  description: Sets the maximum number of items to return in the response.
                  serializedName: maxresults
              protocol: !<!Protocols> 
                http: !<!HttpParameter> 
                  in: header
```

Which Model4 build a `groupBy` parameter with `group name`+`operation name`Options (what specified in swagger as postfix).

Since we already have an Option class for each methods,   Autorest Swift just combine all properties into 1 Option object.

src/AutorestSwift/View Models/OperationViewModel.swift
sacheu: This check is needed for building signature of the operation, not for creating the model file.

This is needed as there are scenario where

Like in Paging, it has this `GroupProperty`:
Technically, it should be folded into method, Option object. However, the parameter `offset` is in `path`, so that should not be part of the Option object. So we should put that in signature parameters , like all other `path` parameters.
```
          originalParameter:
            - !<!Parameter> &ref_44
              schema: *ref_2
              groupedBy: *ref_9
              implementation: Method
              required: true
              language: !<!Languages> 
                default:
                  name: offset
                  description: Offset of return value
                  serializedName: offset
              protocol: !<!Protocols> 
                http: !<!HttpParameter> 
                  in: path
```

src/AutorestSwift/View Models/ClientMethodOptionsViewModel.swift
```diff
@@ -53,8 +53,25 @@ struct ClientMethodOptionsViewModel {
         }
         var properties = [PropertyViewModel]()
 
-        parameters.forEach {
-            properties.append(PropertyViewModel(from: $0.value))
+        // In generated code stencil, there is already a 'clientRequestId', so do not duplicate by adding anyone with same name
+        for parameter in parameters where parameter.value.name != "clientRequestId" {
+            properties.append(PropertyViewModel(from: parameter.value))
+        }
```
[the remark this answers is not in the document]
sacheu: Add a todo

src/AutorestSwift/View Models/ClientMethodOptionsViewModel.swift
tjprescott: Would it not make sense to just deconflict the names?

src/AutorestSwift/View Models/ObjectViewModel.swift
tjprescott: In that case, what are the other parameters? Sometimes they are duplicates of `first` but, at least in storage, there were multiple, different values, in which case we would need to figure out what the right answer should be.

src/AutorestSwift/View Models/ClientMethodOptionsViewModel.swift
sacheu: technically, the names are not conflict, it just results in 2 `options` objects  with similar names. That is why I merge the 2 properties into 1 Options

image

src/AutorestSwift/View Models/ObjectViewModel.swift
sacheu: They look duplicate to me. But are 'apiVersion' in the case of paging.json

src/AutorestSwift/View Models/ClientMethodOptionsViewModel.swift
sacheu: And even if we don't merge the pagingGetMultiplePagesOptions into the existing opptions, we still need to add code to generate a model for 'pagingGetMultiplePagesOptions'.

tjprescott (approved):

PR #284, opened 2020-12-02
Add coverage report upload step to CI pipeline

daviwil:
This adds an extra step to the CI pipeline to push testserver coverage numbers to the AutoRest coverage dashboard.  I'll set up a private pipeline with the necessary variables to do this successfully when run on non PR builds on `master`.  I believe the step still passes if the credentials aren't available.

### DIFF
--- a/eng/pipelines/ci.yml
+++ b/eng/pipelines/ci.yml
@@ -17,4 +17,5 @@ jobs:
       - script: |
           npm run coverage-push -- $(Build.Repository.Name) $(Build.SourceBranch) $(github-token) $(storage-coverage-user) $(storage-coverage-pass)
         workingDirectory: node_modules/@microsoft.azure/autorest.testserver
+        condition: and(ne(variables['Build.Reason'], 'PullRequest'), eq(variables['System.TeamProject'], 'internal'))
         displayName: "Upload Coverage Report"
--- a/eng/pipelines/ci.yml
+++ b/eng/pipelines/ci.yml
@@ -13,3 +13,8 @@ jobs:
 
       - script: npm run test
         displayName: Run Tests
+
+      - script: |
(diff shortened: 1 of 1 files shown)


eng/pipelines/ci.yml
```diff
@@ -13,3 +13,8 @@ jobs:
 
       - script: npm run test
         displayName: Run Tests
+
+      - script: |
+          npm run coverage-push -- $(Build.Repository.Name) $(Build.SourceBranch) $(github-token) $(storage-coverage-user) $(storage-coverage-pass)
```
sacheu, line 18: In package.json, we don't have `coverage-push`, do we need to add that before this CI change work?

sacheu (commented):
I have a question about the change.

daviwil:
I was wrong, the coverage step fails when the variables aren't present.  I'll have to see how the other language CI pipelines are getting around this.

eng/pipelines/ci.yml
daviwil: No, it's provided by `autorest.typescript`, you can see the script step has a `workingDirectory` set to that folder path so the `npm` script is coming from there instead.

sacheu: In  https://github.com/Azure/autorest.swift/blob/master/AutorestSwiftTest/ZZZAutoRestReportTest.swift,  we print out the coverage values like:

is this the right format that the coverage site is expecting?


```
        let coverage: Float = Float(passedCount) / Float(totalCount) * 100

        print("Passed Test=\(passedCount), Total Test=\(totalCount), Coverage=\(String(format: "%.2f", coverage))%")
```

sacheu (approved):

daviwil:
I have a separate CI pipeline set up for "internal" CI now which will upload the coverage numbers.  I'll merge this and give it a shot now!

PR #285, opened 2020-12-10
Refactor Autorest.Swift for changes to AzureCore

tjprescott:
Leaves out the changes for ModelFlattening. Those will be addressed in a separate PR.
https://github.com/Azure/azure-sdk-for-ios/pull/596

To facilitate review, the regenerated test code has been placed in PR #286.


src/AutorestSwift/Models/Complex/Schema.swift
```diff
@@ -100,15 +101,31 @@ class Schema: Codable, LanguageShortcut {
...
+            } else {
+                fatalError("Type .dateTime but not schema `DateTimeSchema`")
+            }
+        case .date:
+            swiftType = "SimpleDate"
+        case .unixTime:
+            swiftType = "UnixTime"
         case .byteArray:
-            swiftType = "Data"
+            if (self as? ByteArraySchema)?.format == .base64url {
```
sacheu, line 120: since both cases set swiftType to Data, we can remove the if check here

src/AutorestSwift/Models/Complex/Schema.swift
```diff
@@ -100,15 +101,31 @@ class Schema: Codable, LanguageShortcut {
...
+                case .dateTime:
+                    swiftType = "Iso8601Date"
+                case .dateTimeRfc1123:
+                    swiftType = "Rfc1123Date"
+                }
+            } else {
+                fatalError("Type .dateTime but not schema `DateTimeSchema`")
+            }
+        case .date:
+            swiftType = "SimpleDate"
```
sacheu, line 116: The name 'SimpleDate' seem confusing when there are `Iso88601Date', 'Rfc1123Date' which shows what standard they implement.

src/AutorestSwift/View Models/OperationViewModel.swift
```diff
@@ -118,73 +91,54 @@ struct OperationParameters {
...
-    case date
-    case dateTime
-    case dateTimeRfc1123
     case time
 }
 
 struct BodyParams {
-    let param: ParameterViewModel
+    /// The `ParameterViewModel` corresponding to the body param
+    var param: ParameterViewModel
```
sacheu, line 127: can this bet 'let' instead of 'var'?

src/AutorestSwift/View Models/ResponseViewModel.swift
```diff
@@ -56,42 +52,37 @@ enum ResponseBodyType: String {
...
         } else if type == .byteArray {
             return .byteArrayBody
         } else if input == "[Date]" {
-            if let arraySchema = schema as? ArraySchema,
-                let dateTimeSchema = arraySchema.elementType as? DateTimeSchema {
-                return (dateTimeSchema.format == .dateTimeRfc1123) ? .dateTimeRfc1123Body : .dateTimeBody
-            } else {
-                return .dateBody
-            }
+            return .plainBody
```
sacheu, line 61: we don't need this if check since the else is .plainBody

sacheu (approved):
Look great to me. 

Just a few comments/questions.

And I remember in paging.json, it uses x-ms-parameter-grouping" with Options as suffix.
Does AutoRest Swift generates a new class for this now? Or it will merge properties in there to the MethodOption that it always generated ?


          {
            "name": "maxresults",
            "in": "header",
            "required": false,
            "type": "integer",
            "format": "int32",
            "description": "Sets the maximum number of items to return in the response.",
            "x-ms-parameter-grouping": {
                "postfix": "Options"
            }
          },

sacheu:
And I recall you mention refactoring the generated code to use `error_map` like you saw in python. Is that still part of the plan?

tjprescott:
> Look great to me.
> 
> Just a few comments/questions.
> 
> And I remember in paging.json, it uses x-ms-parameter-grouping" with Options as suffix.
> Does AutoRest Swift generates a new class for this now? Or it will merge properties in there to the MethodOption that it always generated ?
> 
> ```
>       {
>         "name": "maxresults",
>         "in": "header",
>         "required": false,
>         "type": "integer",
>         "format": "int32",
>         "description": "Sets the maximum number of items to return in the response.",
>         "x-ms-parameter-grouping": {
>             "postfix": "Options"
>         }
>       },
> ```

I preserved the existing behavior where they are merging into the method options and left a TODO to revisit this.

tjprescott:
> And I recall you mention refactoring the generated code to use `error_map` like you saw in python (to make the generated code less verbose). Is that still part of the plan?

That's on my TODO list, but the current PR doesn't address that.

src/AutorestSwift/Models/Complex/Schema.swift
tjprescott: True. I just didn't know a good format name for YMD. I also considered whether to make all these date types actually return their value (and thus type Date) and simply use these internally.

src/AutorestSwift/View Models/OperationViewModel.swift
tjprescott: No because OperationParam can update the name. This was to get rid of an initializer that had an optional `withName` in the signature.

Issue #299, opened 2021-02-05
Generated enum case has difficulty with certain keywords


JoshuaLai:
![Screen Shot 2021-02-05 at 1 18 52 PM](https://user-images.githubusercontent.com/9044372/107073032-b178ae00-679b-11eb-8825-5c8763440a7a.png)

As per screenshot. The enum cases generated from autorest doesn't seem to add the ` `` ` around keywords such as `public`.

tjprescott:
Fixed by #300.

PR #300, opened 2021-02-11
Backtick values which correspond to Swift reserved keywords

tjprescott:
Fix issue #299.

cc/ @JoshuaLai


src/AutorestSwift/Utils/ReservedKeywords.swift
```diff
@@ -0,0 +1,105 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+private let reservedTypes = [
+    "Error": true,
```
JoshuaLai, line 30: Every thing looks good just not sure do we care about the Objective-C types? Or burn that bridge when we get there...if we get there...

src/AutorestSwift/Utils/ReservedKeywords.swift
```diff
@@ -0,0 +1,105 @@
...
+    "throw": true,
+    "throws": true,
+    "true": true,
+    "try": true
+]
+
+extension String {
+    /// Returns `true` if the string corresponds to a reserved type
+    var isReservedType: Bool {
+        return reservedTypes[self] ?? false
```
JoshuaLai, line 98: Do we worry about case sensitivity. Should we do `self.lowercase`?

tjprescott: In this case, no, because Swift keywords are case-sensitive.

src/AutorestSwift/Utils/ReservedKeywords.swift
tjprescott: This isn't even an exhaustive list of Swift types so I'll opt for the when we get there (if we get there). Since ObjC has no namespaces or modules and uses the prefix convention, the chance of collision with ObjC should be small.

tjprescott: Also, I'm going to switch the implementation to use Swift Sets. The efficiency of the dictionary without the throwaway boolean values.

JoshuaLai (approved):

sacheu (approved):
Great change, Travis. It covers a lot more swift keywords.

PR #309, opened 2021-03-09
Load autorest command line args

tjprescott:
Previously, we only loaded a single argument "packageName", but really, we should be loading all supported autorest arguments. This will allow us to customize code generation based on command line args and/or use of the README.md file.


src/AutorestSwift/JsonRpc/AutorestPlugin.swift
```diff
@@ -36,8 +36,7 @@ class AutorestPlugin {
     var processRequestId: Int!
 
     // TODO: Increase when working correctly
-    private let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
-    private let group = DispatchGroup()
+    var eventLoop = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
sacheu, line 39: Why do we need need more than 1 thread? even when working correctly?

src/AutorestSwift/Code Generator/Manager.swift
```diff
@@ -33,56 +33,68 @@ struct StringDecodeKeys: ExtensionStringDecodeKeys {
...
     }
 
     /// Initialize with a raw YAML string (i.e. the way it is received from Autorest).
-    init(withString string: String, packageName: String? = nil) {
-        self.mode = .autorest
-        self.inputString = string
+    func configure(input: String, client: ChannelClient, sessionId: String, completion: @escaping () -> Void) {
+        guard configured == false else {
+            SharedLogger.warn("Manager is already configured.")
+            return
```
sacheu, line 82: Should this also call `fatalError` when configured is true, same as what you did on line 96

src/AutorestSwift/Code Generator/Manager.swift
```diff
@@ -33,56 +33,68 @@ struct StringDecodeKeys: ExtensionStringDecodeKeys {
...
-    }
+    static var shared = Manager()
 
     // MARK: Properties
 
-    let inputString: String
-
-    let destinationRootUrl: URL
+    var configured: Bool
+    var config: ManagerConfig
```
sacheu, line 41: I wonder if it is a good idea to change config to ManagerConfig? where `nil` means it is not configured and not nil mean it is configured. That way, we don't need to maintain an extra flag `configured` to indicates whether it is configured or not.

src/AutorestSwift/Code Generator/Manager.swift
```diff
@@ -33,56 +33,68 @@ struct StringDecodeKeys: ExtensionStringDecodeKeys {
...
+    var inputString: String
+    var destinationRootUrl: URL!
     var packageUrl: URL?
```
sacheu, line 44: Should destinationRootUrl and packageUrl part of ManagerConfig as well?

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
```diff
@@ -165,22 +165,16 @@ class AutorestPlugin {
...
+            encoder.outputFormatting = .prettyPrinted
+            let encoded = try! encoder.encode(Manager.shared.config)
+            let arguments = String(data: encoded, encoding: .utf8)!
+            SharedLogger.info("Autorest Configuration: \(arguments)")
             do {
-                try manager.run()
+                try Manager.shared.run()
 
-                guard let packageUrl = manager.packageUrl else {
+                guard let packageUrl = Manager.shared.packageUrl else {
```
sacheu, line 177: Why we do this check (check if packageUrl is nil) here (in line 177) after manager is run()? shouldn't we check this before Manager.run?

src/AutorestSwift/View Models/PackageFileViewModel.swift
```diff
@@ -29,13 +29,13 @@ import Foundation
...
+    let version: String
 
     init(from model: CodeModel) {
-        // FIXME: Blunt-instrument to override the package name from the code model
-        let commandLineArguments = CommandLineArguments()
-        self.name = Manager.packageName ?? commandLineArguments["--package-name"] ?? model.packageName
+        let config = Manager.shared.config
+        self.name = config.packageName ?? model.packageName
+        self.namespace = config.namespace ?? ""
+        self.version = config.packageVersion ?? ""
```
sacheu, line 39: Should we failed the code generation if there is no values for 'namespace' and 'version'? instead of setting them to empty string and continue?

src/AutorestSwift/Code Generator/ManagerConfig.swift
```diff
@@ -0,0 +1,160 @@
...
+    /// - Parameter sessionId: RPC session ID
+    func loadValues(client: ChannelClient, sessionId: String, completion: @escaping () -> Void) {
+        self.client = client
+        self.sessionId = sessionId
+        // FIXME: Clearly, this is garbage, but I can't force these to behave synchronously
+        get(value: "input-filename") {
+            self.inputFilename = $0
+            self.get(value: "output-directory") {
+                self.outputDirectory = $0
+                self.get(value: "clear-output-folder") {
```
sacheu, line 93: Comment: This block is a bit hard to read. I understand each parameters need to be retrieve asynchronously.
I wonder if we can have an array of all the parameters we need to retrieve and maintains an index of where in the array that we have already retrieve the values. So in the call back of the get function, it get the next item of the array and call get again. If it reaches the end, call completion().

sacheu (approved):
Looks good to me. Just have a few suggestions/questions.

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
tjprescott: I haven't tried changing this to see, but it could speed things up a bit when blocking I/O is happening like writing files.

src/AutorestSwift/Code Generator/Manager.swift
tjprescott: If it's already configured, this just warns you and lets you continue with how it was previously configured. However, if you try to run autorest without configuring it, there's no possible way it will work, which is why that is a fatal error.

src/AutorestSwift/Code Generator/ManagerConfig.swift
tjprescott: I tried many ways to fix this ugly chaining and they all ended up in thread limbo. I'll give this a shot, though I may need to switch from attribute syntax to dictionary to make it work.

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
tjprescott: I'll have to check, but I think packageUrl is set during `run`.

src/AutorestSwift/View Models/PackageFileViewModel.swift
tjprescott: Possibly. We were hard-coding them to "secret" and "0.1" so I don't think these values are currently used anywhere in our codegen.

src/AutorestSwift/Code Generator/Manager.swift
tjprescott: Great point

src/AutorestSwift/Code Generator/Manager.swift
tjprescott: Probably

src/AutorestSwift/JsonRpc/AutorestPlugin.swift
tjprescott: Yes, this is set during `run()`.

PR #312, opened 2021-04-07
Fix issues with AutoRest Jazzy generation

tjprescott:
- uses xcodebuild instead of `swift build` since the latter only works on macOS.
- generated jazzy files in the `jazzy` folder instead of `.jazzy`.

### DIFF
--- a/templates/Jazzy_File.stencil
+++ b/templates/Jazzy_File.stencil
@@ -10,7 +10,6 @@ output: ../build/jazzy/{{ model.name }}
 swift_build_tool: spm
 clean: true
 sdk: iphonesimulator
-sdk: iphonesimulator
 xcodebuild_arguments:
   - "-workspace"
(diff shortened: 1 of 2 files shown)


templates/Jazzy_File.stencil
```diff
@@ -10,12 +10,16 @@ output: ../build/jazzy/{{ model.name }}
 swift_build_tool: spm
 clean: true
 sdk: iphonesimulator
-swift_build_tool: spm
-build_tool_arguments:
-  - "-Xswiftc"
-  - "-swift-version"
-  - "-Xswiftc"
-  - "5"
+sdk: iphonesimulator
```
JoshuaLai, line 13:
```suggestion
```
This line is duplicated

JoshuaLai (changes requested):
Just on tiny thing to clean up

JoshuaLai (approved):

PR #319, opened 2021-08-12
Invoke CompletionHandler as .failure if status code is not legit.

PotterDai:
Currently generated client won't invoke CompletionHandler if the status code is not legit.


PotterDai:
Close since the bug lies deeper in the AzureCore

